PR #5, opened 2016-02-10
Update duet.ml

trucnguyenlam:
make compiling work


zkincaid:
Thanks for the patch -- this line was intended for a different (experimental) branch.

PR #6, opened 2017-02-02
Improvements to PA emptiness check

tm507211:
Improves the PA emptiness check. Implements monadic embedding queries as bipartite maximum graph matching. Also implements a search tree for a directed search for a covering configuration.

### DIFF
--- a/pa/biGraph.ml
+++ b/pa/biGraph.ml
@@ -1,0 +1,176 @@
+open Pervasives
+open BatPervasives
+open Apak
+
+include Log.Make(struct let name = "BiGraph" end)
+
+module type Vertex = sig
+  type t
+  val hash : t -> int
+  val equal : t -> t -> bool
+  val compare : t -> t -> int
+  val pp : Format.formatter -> t -> unit
+end
+
+module type S = sig
+  type t
+  type vertex
+
+  val pp : Format.formatter -> t -> unit
+
+  val empty : t
+  val make : vertex BatEnum.t -> vertex BatEnum.t -> (vertex * vertex) BatEnum.t -> t
+  val max_matching : t -> int
+  val u_size : t -> int
+  val v_size : t -> int
+  val e_size : t -> int
+end
+
+module Make (V : Vertex) = struct
+  type vertex = V.t
+  type edge = V.t * V.t [@@deriving ord]
+
+  module VertexSet = BatSet.Make(V)
+  module EdgeSet = BatSet.Make(struct
+      type t = edge [@@deriving ord]
+    end)
+
+  type t =
+    { u : VertexSet.t;
+      v : VertexSet.t;
+      e : EdgeSet.t }
+
+  let pp_edge formatter (u, v) =
+    Format.fprintf formatter "(%a, %a)" V.pp u V.pp v
+
+  let pp formatter graph =
+    Format.fprintf formatter "{U: {%a}, V: {%a}, E: {%a}}"
+    (ApakEnum.pp_print_enum V.pp) (VertexSet.enum graph.u)
+    (ApakEnum.pp_print_enum V.pp) (VertexSet.enum graph.v)
+    (ApakEnum.pp_print_enum pp_edge) (EdgeSet.enum graph.e)
+
+  let empty = 
+    { u = VertexSet.empty;
+      v = VertexSet.empty;
+      e = EdgeSet.empty }
+
+  let make u v e =
+    { u = VertexSet.of_enum u;
+      v = VertexSet.of_enum v;
+      e = EdgeSet.of_enum e}
+
+  module VMap = BatMap.Make(V)
+  module VoptMap = BatMap.Make(struct
+    type t = V.t option
+    let compare x y = (* None is the least element *)
+      match (x, y) with
+        None, None -> 0
+      | None, Some _ -> -1
+      | Some _, None -> 1
+      | Some x', Some y' -> compare x' y'
+  end)
+
+(*  let make_adj e =
+    let f (u, v) map =
+      VMap.add u v map
+    in
+    EdgeSet.fold f e VMap.empty *)
+
+  let make_adj e =
+    let f (u, v) (uMap, vMap) =
+      let g x y map =
+        if (VMap.mem x map) then
+          VertexSet.add y (VMap.find x map)
+        else
+          VertexSet.add y VertexSet.empty
+      in
+      ((VMap.add u (g u v uMap) uMap), (VMap.add v (g v u vMap) vMap))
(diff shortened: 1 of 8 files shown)


pa/biGraph.ml
```diff
@@ -0,0 +1,176 @@
+open Pervasives
```
zkincaid: "matching" might be a more appropriate name than "bigraph" for this file (following the convention that <module>.Make is a functor that makes a <module> -- BiGraph.Make makes an implementation of the matching algorithm).

pa/biGraph.ml
```diff
@@ -0,0 +1,176 @@
...
+      v = VertexSet.empty;
+      e = EdgeSet.empty }
+
+  let make u v e =
+    { u = VertexSet.of_enum u;
+      v = VertexSet.of_enum v;
+      e = EdgeSet.of_enum e}
+
+  module VMap = BatMap.Make(V)
+  module VoptMap = BatMap.Make(struct
```
zkincaid: 1. Is it important that None is the least element?  (If not we use [@@deriving ord] to implement the comparison).
2. I find the use of VoptMap to be a little confusing.  What's the advantage of using a VoptMap instead of a pair consisting of a VMap plus an extra element?

pa/biGraph.ml
```diff
@@ -0,0 +1,176 @@
...
+module Make (V : Vertex) = struct
+  type vertex = V.t
+  type edge = V.t * V.t [@@deriving ord]
+
+  module VertexSet = BatSet.Make(V)
+  module EdgeSet = BatSet.Make(struct
+      type t = edge [@@deriving ord]
+    end)
+
+  type t =
```
zkincaid: This representation isn't really used: in max_matching you immediately compute an adjacency map representation.  No need to store this representation.

pa/biGraph.ml
```diff
@@ -0,0 +1,176 @@
...
+        if (VMap.mem x map) then
+          VertexSet.add y (VMap.find x map)
+        else
+          VertexSet.add y VertexSet.empty
+      in
+      ((VMap.add u (g u v uMap) uMap), (VMap.add v (g v u vMap) vMap))
+    in
+    EdgeSet.fold f e (VMap.empty, VMap.empty)
+
+  let bfs u adj_u pair_u pair_v =
```
zkincaid: Comment needed

pa/searchTree.ml
```diff
@@ -0,0 +1,132 @@
...
+  type baseSet
+  type elt
+
+  val empty : baseSet -> (elt -> baseSet) -> t
+  val make  : baseSet -> (elt -> baseSet) -> elt BatSet.t -> t
+  val insert : t -> elt -> unit
+  val covered : (elt -> elt -> bool) -> t -> elt -> elt option
+end
+
+module Make (Base : Element) (Elt : Element) = struct
```
zkincaid: Comment needed.  In particular, it's not clear what Base and Elt are.

pa/searchTree.ml
```diff
@@ -0,0 +1,132 @@
...
+  type tree = (* Every node is a set of elements *)
+    Leaf
+  | Node of node
+  and node =
+    { mutable data : EltSet.t;
+      left : tree;
+      right : tree }
+
+  type t =
+    { project : elt -> baseSet;
```
zkincaid, line 56: Is "project" the right word?

pa/searchTree.ml
```diff
@@ -0,0 +1,132 @@
...
+    BatSet.iter f items;
+    tree
+
+  (* This performs the actual search
+     returns true if there exists a elt in tree
+     such that (f elt item) *)
+  let covered f stree item =
+    let neg : int ref = ref 0 in
+    let pos : int ref = ref 0 in
+    let inc x = x := !x + 1 in
```
zkincaid: this is the build in function "incr"

pa/searchTree.ml
```diff
@@ -0,0 +1,132 @@
...
+               | e :: elist ->
+                  if i = e then
+                    match go elist iblist' left with
+                      None -> (go elist iblist' right)
+                    | Some x -> Some x
+                  else
+                    (go elist iblist left)
+          end
+    in let ret = go elist iblist stree.tree in
+       (*       print_int !neg; print_string " "; print_int !pos; print_newline(); *)
```
zkincaid: Ocaml's standard library includes a Printf module

pa/searchTree.ml
```diff
@@ -0,0 +1,132 @@
...
+    let f item =
+      insert tree item
+    in
+    BatSet.iter f items;
+    tree
+
+  (* This performs the actual search
+     returns true if there exists a elt in tree
+     such that (f elt item) *)
+  let covered f stree item =
```
zkincaid: Comment: only works if "f elt item => (project elt) is a subset of (project item)".  Since f returns a bool, we conventionally use "p" instead of "f".

pa/struct.ml
```diff
@@ -263,9 +266,9 @@ module Make (P : Symbol) = struct
 
   (* Is there an embedding (injective homomorphism) of x into y? *)
   let embeds x y =
```
zkincaid: maybe rename to "embeds_naive"?

pa/biGraph.ml
```diff
@@ -0,0 +1,176 @@
...
+
+module type Vertex = sig
+  type t
+  val hash : t -> int
+  val equal : t -> t -> bool
+  val compare : t -> t -> int
+  val pp : Format.formatter -> t -> unit
+end
+
+module type S = sig
```
zkincaid: This signature is unused.  If you want Make to be associated with this signature, the best way to do it is to create a separate .mli file (and add comments for top-level functions to this file).

pa/struct.ml
```diff
@@ -286,7 +289,80 @@ module Make (P : Symbol) = struct
...
+      let g ids id =
+        KSet.add id ids
+      in
+      List.fold_left g ids args
+    in
+    AtomSet.fold f str.prop KSet.empty
+
+  module Graph = BiGraph.Make(Putil.PInt)
+
+  let mk_edges x_ids x_sigs y_ids y_sigs =
```
zkincaid: move inside mk_graph?

zkincaid (changes requested):
Just a few style comments.

pa/biGraph.ml
tm507211: It doesn't matter that None is the least element. VoptMap was a way of allowing None and Some values to be treated the same in the code for DFS. It should be possible to change it to be a pair of a VMap and another element.

PR #7, opened 2017-06-05
Embeds

tm507211:
Updates embedding algorithm to be more modular and adds "arc consistency" code.


pa/cc/matching_stubs.c
```diff
@@ -284,267 +160,147 @@ bool embedding(const vector<vector<uint8_t>>& sigs1, const vector<vector<uint8_t
...
+  while (!q.empty()){
+    size_t pu = q.top().second; q.pop();
+    vector<Graph::Edge> adj = p_graph.uAdj(pu);
+    for (size_t i = 0; i < adj.size(); ++i){
+      p_removed.clear(); u_removed.clear();
+      emb.choose_constraint(pu, adj[i].vertex, p_removed, u_removed);
+      if (!emb.get_valid()){
+	emb.add_back(p_removed, u_removed);
+      } else {
+	decisions.emplace(pu, i, u_removed, p_removed, adj);
```
zkincaid, line 613: Return true here?

This code seems to eagerly make as many decisions about *all* conflicts, rather than just choosing *a* conflict to make a decision.  This subverts the consistency check that every decision must be consistent with some maximal matching (search might descend deep into a subtree that could have been pruned early because it can't be extended to a maximal matching).

pa/cc/embedding.h
```diff
@@ -0,0 +1,272 @@
...
+	valid_ = false;
+	break;
+      }
+      pu_units.clear(); pv_units.clear();
+      //filter(uu_units, uv_units, p_removed, pu_units, pv_units);
+    }
+  }
+
+  /* Remove any edge inconsistent with U[i] |-> V[i] in the universe graph */
+  void filter(const std::vector<int>& U, const std::vector<int>& V,
```
zkincaid: This filtering mechanism is ineffective, according to my limited experimentation -- it doesn't ever remove any edges in the examples I've tried (after uncommenting line 224).  I haven't tracked down why.

pa/cc/graph.h
```diff
@@ -0,0 +1,316 @@
...
+      ans += (matches_u[i] != -1) || dfs(matches_u, matches_v, vis, i, i);
+    }
+    return ans;
+  }
+
+  bool unit_prop(std::vector<VertexPair>& removed, std::vector<int>& u_units, std::vector<int>& v_units){
+    std::queue<size_t> units;
+    for (size_t i = 1; i < adj_u.size(); ++i){
+      if (adj_u[i].size() == 0) return false;
+      if (adj_u[i].size() == 1 && adj_v[adj_u[i][0].vertex].size() != 1) units.push(i);
```
zkincaid, line 99: Why the guard `adj_v[adj_u[i][0].vertex].size() != 1`?  (Similarly for line 115)

pa/cc/graph.h
```diff
@@ -0,0 +1,316 @@
...
+    }
+    return true;
+  }
+
+  /* Remove a single edge
+     Assumption:
+       ajd_u[u][pos] = {v, pos'} <->
+       adj_v[v][pos'] = {u, pos}
+   */
+  void remove_edge(size_t u, size_t pos){
```
zkincaid, line 134: I think the second argument being a *position* rather than a vertex is a confusing interface.  It's nice that you get a constant-time removal operation, but is this a bottleneck?

pa/cc/graph.h
```diff
@@ -0,0 +1,316 @@
...
+    }
+    adj_u[u].pop_back();
+    adj_v[k.vertex].pop_back();
+  }
+
+  /* Remove all edges inconsistent with U[i] |-> V[i],
+     Assumptions:
+       U.size() == V.size()
+   */
+  std::vector<VertexPair> remove_edges(const std::vector<int>& U, const std::vector<int>& V){
```
zkincaid: I suggest renaming to "commit_edges" or something -- without reading the comment, I would have guessed that this function removes the edges passed as its parameters.

zkincaid (approved):

pa/cc/matching_stubs.c
zkincaid: "here" means after line 280.

tm507211: Yes, "return true" should be there.

pa/cc/graph.h
tm507211: The reasoning is that I only wanted to capture "new units." That way when one inspects the values of u_units and v_units it only consists of edges unit propagated this time instead of in previous calls of the unit_propagation (or from some other means).

pa/cc/graph.h
tm507211: I'm not sure, if it's the bottleneck. The arguments are basically an edge,
maybe the better way to specify it is `void remove_edge(Edge e){ ...`

pa/cc/graph.h
tm507211: This makes sense.

PR #10, opened 2019-01-24
use a single set of variables for the return variables

aeflores:
This avoids generating so many different variables which helps
with performance.


aeflores:
I could not find an implementation of **take** in the standard library though I am sure it is somewhere. I would be happy to learn where!

PR #13, opened 2019-08-15
termination reasoning

happypig95:
current .merlin does not work well for me (I'm using merlin+VS Code on mac) especially for the ppx libraries. I have used a script that invokes ocamlfind and automatically add the paths of all libraries into the the merlin file but the paths are local so it might be a good idea not to include that file here. 
In the near future there might be changes to the test suites and the regression tests as well.


AUTHORS.txt
```diff
@@ -0,0 +1,8 @@
+(* OASIS_START *)
```
zkincaid: Remove AUTHORS.txt

apak/extGraph.ml
```diff
@@ -47,7 +47,7 @@ module type Vertex = sig
   module Set : Putil.Hashed.Set.S with type elt = t
 end
 
-let display_image = ref "eog"
+let display_image = ref "qlmanage -p "
```
zkincaid: Don't change the default previewer, but could make it configurable by a command line argument

duet/cra.ml
```diff
@@ -610,6 +612,30 @@ let analyze file =
...
+  
+
+let prove_termination_main file =
+  populate_offset_table file;
+  match file.entry_points with
+  | [main] -> begin
+      let rg = Interproc.make_recgraph file in
+      let entry = (RG.block_entry rg main).did in
+      let (ts, assertions) = make_transition_system rg in
+      TSDisplay.display ts;
```
zkincaid: Should only call display if CmdLine.display_graphs is set

output
```diff
@@ -0,0 +1,643 @@
+= Parsing ======================================================================
```
zkincaid: Remove output

test/concurrent/README.txt
```diff
@@ -1,3 +0,0 @@
-This directory contains simplified benchmarks from the SV-COMP 17 concurrency
```
zkincaid: Do not delete

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+        loop_body_vertices: vertex list;
+        back_edge: vertex * vertex * 'a;
+        children: 'a loops list;
+        depth: int;
+        header_f: 'a;
+        body_f: 'a;
+    }
+
+let print_loop {header; splitted_hd; loop_body_vertices; back_edge; children; depth; header_f; body_f} =
+    printf "\n======== Loop ======== \n";
```
zkincaid: **Throughout**: logf instead of printf for printing debugging info

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+        (fun (u, weight, v) (acc) -> (f u <> i_th_scc) || acc)
+        wg
+        v
+        false
+    
+(** find a header and a backedge that goes into the header *)
+(* replace filter with find *)
+let find_header_backedge wg f i_th_scc scc =
+    printf "... finding a header and a backedge ... \n";
+    let headers = List.filter (is_header wg f i_th_scc) scc in
```
zkincaid: Replace filter with find

srk/src/weightedGraph.ml
```diff
@@ -29,6 +30,10 @@ type 'a t = 'a weighted_graph
 
 type vertex = int
 
+let get_algebra wg = wg.algebra
+
+let get_true_from_algebra wg = wg.algebra.one
```
zkincaid: This doesn't seem necessary -- just use (get_algebra wg).one.
Also, 1 does not represent "true" (thinking of weights as formulas) -- it represents the identity relation.

srk/src/weightedGraph.ml
```diff
@@ -55,6 +60,27 @@ let add_edge wg u weight v =
     { wg with graph = U.add_edge wg.graph u v;
               labels = M.add (u, v) weight wg.labels }
 
+let get_non_trivial_scc wg =
```
zkincaid: Careful: graphs can have self loops -- I would call a self-loop a non-trivial SCC, but it has size one.

srk/src/weightedGraph.ml
```diff
@@ -103,6 +136,8 @@ let contract_vertex wg v =
 
 let max_vertex wg = U.fold_vertex max wg.graph 0
 
+let max_vertex_plus_1 wg = U.fold_vertex max wg.graph 1
```
zkincaid: This reads like it should compute (max_vertex wg) + 1, but it doesn't do that

srk/src/weightedGraph.mli
```diff
@@ -15,7 +15,8 @@ type 'a algebra =
     zero : 'a;
     one : 'a }
 
-(** Unweighted graphs *)
```
zkincaid, line 18: Don't delete this comment.  The (** Unweighted graphs *) comment is attached to the U module, which is used to access the underlying graph structure that we get by forgetting about the weights on a weighted graph.

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+   Format.fprintf Format.std_formatter "\nLooking for id of symbol:\n"; 
+   pp_symbol srk Format.std_formatter symbol;  *)
+  let tid = CoordinateSystem.cs_term_id cs (`App (symbol, [])) in
+    (* Format.fprintf Format.std_formatter "\nid for this symbol:\n"; 
+    Format.fprintf Format.std_formatter "%d\n" tid;  *)
+    let c = Vec.coeff tid vec in
+       (* QQ.pp Format.std_formatter c;  *)
+      c
+
+let get_coeffs_from_eq_as_if_leq0 cs eq_t list_symbols =
```
zkincaid, line 87: This seems to serve the purpose of changing the vector eq_t from the coordinate system cs to the one defined by list_symbols.  Why do we want to do this?

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+            let row2 = get_coeffs_from_eq_as_if_leq0 cs t list_symbols in
+            let neg_row2 = Vec.negate row2 in
+            let m = Mat.add_row (i+1) neg_row2 m1 in
+            (m, i+2)
+          end
+    )
+    (Mat.zero, 0)
+    constraints
+
+let rec create_const_symbols srk name n =
```
zkincaid: (1 -- n) /@ (fun _ -> mk_const srk (mk_symbol srk ~name:name `TyReal))
|> BatEnum.of_list

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+include Log.Make(struct let name = "srk.nestedLoops" end)
+
+(* module SCC = Graph.Components.Make(U) *)
+module Solver = SrkZ3.Solver
+
+(* TODO: refactor loop nesting relation from label algebras *)
+type 'a loops =
+    { 
+        header: vertex;
+        splitted_hd: vertex;
```
zkincaid: What are the splitted_hd, header_f, and body_f fields?
Do we need to track depth?
Is a *single* back-edge sufficient?

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+            wg
+            hd
+            []
+        in
+        match backedge_list with
+        | [] -> assert false
+        | t :: _ -> (fst t, hd, snd t)
+
+(** replace the pos-th element in l with a *)
+let replace l pos a =
```
zkincaid, line 62: BatList.modify_at

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+                    (assign_child_loops l_parent tail f) 
+                    parent_ord 
+                    (Some { parent_loop with children = hd :: parent_loop.children })
+                    
+
+(** Remove all backedges (top-level) recorded in a list of loops from wg *)
+let rec remove_all_backedges wg l_loops =
+    match l_loops with
+    | [] -> wg
+    | loop_op :: tail ->
```
zkincaid: Style: collapse nested pattern matching:
```
match l_loops with
    | [] -> wg
    | None :: tail -> remove_all_backedges wg tail
    | Some loop :: tail -> ...
```

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+    let s = mk_const srk (mk_symbol srk ~name:name `TyReal) in
+      s :: create_const_symbols srk name (n-1)
+
+let build_formula srk lambdas mat n_rows n_cols op =
+  let m = Mat.transpose mat in
+  let formulas = ref (mk_true srk) in
+  for i = 0 to n_cols-1 do
+    let row_i = Mat.row i m in
+    let lhs = (mk_zero srk) in
+    let rlhs = ref lhs in
```
zkincaid, line 137: QQVector.term_of_vec srk (List.nth lambdas) row_i

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+  !formulas
+
+let build_lambda_mat_prod_terms srk lambdas mat n_rows n_cols =  
+  let m = Mat.transpose mat in
+  let r = ref [] in
+  for i = 0 to n_cols-1 do
+    let row_i = Mat.row i m in
+    let lhs = (mk_zero srk) in
+    let rlhs = ref lhs in
+    for j = 0 to n_rows-1 do
```
zkincaid, line 163: QQVector.term_of_vec srk (List.nth lambdas) row_i

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+      let term_v = mk_real srk ith_ele in
+      let prod = mk_mul srk [ List.nth lambdas j; term_v] in
+      let value = mk_add srk [ prod; !rlhs] in
+      rlhs := value
+    done;
+    r := List.concat [!r; [!rlhs]]
+  done;
+  !r
+
+let build_gt_zero srk lambdas =
```
zkincaid, line 175: let zero = mk_zero srk in
List.map (mk_leq srk zero) lambdas
|> mk_and srk

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+  List.fold_left
+    (fun f lambda ->
+      mk_and srk [f; mk_leq srk (mk_zero srk) lambda]
+    )
+    (mk_true srk)
+    lambdas
+
+let rec build_subtracted_lambdas srk lambdas1 lambdas2 =
```
zkincaid, line 180: List.map2 (mk_sub srk) lambdas1 lamdas2

srk/src/transition.ml
```diff
@@ -438,4 +439,51 @@ struct
     |> Nonlinear.linearize srk
     |> rewrite srk ~down:(nnf_rewriter srk)
     |> Abstract.abstract ~exists srk man
+
+
+
+    let to_transition_formula_with_consts tr additional_symbols_list =
```
zkincaid, line 447: This seems to be unused.  Remove?

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+    for j = 0 to n_rows-1 do
+      let ith_ele = Vec.coeff j row_i in
+      let term_v = mk_real srk ith_ele in
+      let prod = mk_mul srk [ List.nth lambdas j; term_v] in
+      let value = mk_add srk [ prod; !rlhs] in
+      rlhs := value
+    done;
+    let lhs_exp = !rlhs in
+    let formula = match op with
+    | "eq_0" -> mk_eq srk lhs_exp (mk_zero srk)
```
zkincaid: You should use a datatype instead of passing strings around.

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+            (m, i+2)
+          end
+    )
+    (Vec.zero, 0)
+    constraints
+  
+let print_formula_to_stdout srk descrip f =
+  logf descrip;
+  let s2 = (Printf.sprintf "\n%s\n" (Formula.show srk f)) in
+  logf "%s" s2;
```
zkincaid, line 211: logf "@\n%a@\n" (Formula.pp srk) f

See the OCaml Format module for more information about how to use logf & errorf.

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+  
+  logf "\nIt has to decrese by this amount in each iter: %a" QQ.pp (QQ.sub QQ.zero (List.nth d 0)); 
+
+  (* let t = QQ.negate (List.nth d0 0) in *)
+  let t = QQ.sub (List.nth d 0) (List.nth d0 0) in 
+  logf "\nAnd its lower bound is: %a \n" QQ.pp t;
+  logf "\n\n***************************\n\n";
+  ()
+
+let prove_termination srk tto_transition_formula tmul loop =
```
zkincaid: Parameterizing by tto_transition_formula is a bit weird, and tmul seems to be unused.
I think it would preferable to just take in a "stem" formula, a "loop" formula, and a list of symbol pairs.

zkincaid: Also, the function should return an ranking function (option), not print it:
- There are lots of applications where we want to use termination proving as a sub-procedure
- Even if we only care about termination, when we run on termination benchmarks we want to decide whether the whole program terminates, which means we need to know whether prove_termination succeeds for *every* loop

srk/src/termination.ml
```diff
@@ -0,0 +1,429 @@
...
+
+  let eq4 = build_formula srk lambda2s mat_b n_rows 1 "lt_0" in
+  print_formula_to_stdout srk "\nConstraints eq4:\n" eq4;
+
+  let all_eqs = mk_and srk [l1gt0; l2gt0; eq1; eq2; eq3; eq4] in
+  let rf_terms_list = build_lambda_mat_prod_terms srk lambda2s mat_ap n_rows n_cols in
+  (* List.iter (fun x -> Syntax.Term.pp srk Format.std_formatter x; Format.fprintf Format.std_formatter "\n" ) rf_terms_list; *)
+  let rf_delta0 = build_lambda_mat_prod_terms srk lambda1s mat_b n_rows 1 in
+  let rf_delta = build_lambda_mat_prod_terms srk lambda2s mat_b n_rows 1 in
+  (* let solver = SrkZ3.mk_solver srk in
```
zkincaid, line 386: Generally speaking: strongly prefer not to commit code that's been commented out.

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+        false
+    
+(** find a header and a backedge that goes into the header *)
+(* replace filter with find *)
+let find_header_backedge wg f i_th_scc scc =
+    printf "... finding a header and a backedge ... \n";
+    let headers = List.filter (is_header wg f i_th_scc) scc in
+    match headers with
+    | [] -> assert false
+    (* we assume that only allowing 1 header is enough *)
```
zkincaid: This is a worrying comment... what happens if a loop has >1 header?

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+            in
+            {
+                loop with
+                    header_f = construct_header_guard 
+                                    (path_weight wg_no_bg_and_splitted_headers start splitted_hd) 
+                                    bf;
+                    children = processed_children
+            }
+
+let rec flatten_loops loops_l results =
```
zkincaid, line 258: This is equivalent to
List.rev_map (fun loop -> (loop.header_f, loop.body_f)) loops_l

I think "loops_l" is intended to expand to "loops_list" --- in OCaml, the "s" conventionally designates a list (loops is a list of things, each of which is a loop).

srk/src/nestedLoops.ml
```diff
@@ -0,0 +1,311 @@
...
+            | Some loop ->
+            let (u, v, _) = loop.back_edge in
+            let wgn = remove_edge wg u v in
+                remove_all_backedges wgn tail
+        end
+
+(** compute a tree of nested loops, returns a tuple of
+    (a wg with all backedges deleted, a list of loops at the top level)
+*)
+let rec compute_nested_loops wg d =
```
zkincaid, line 117: It looks like you're computing a tree level-by-level.  I think it would simplify the algorithm if you computed sub-tree by sub-tree (so if the root has k children, you compute the entire sub-tree for the first child, then second, ... through the kth).

srk/test/test_nestedLoops.ml
```diff
@@ -0,0 +1,577 @@
...
+  Format.fprintf Format.std_formatter "\nAnd its lower bound is:\n";
+  let t = QQ.negate (List.nth d0 0) in
+  (* let t = QQ.sub (List.nth d 0) (List.nth d0 0) in *) 
+  QQ.pp Format.std_formatter t;
+  Format.fprintf Format.std_formatter "\n\n";
+
+  Format.fprintf Format.std_formatter "\n\n***************************\n\n";
+  ()
+
+let prove_termination loop =
```
zkincaid, line 357: After you rewrite your prove_termination procedure to return a result rather than printing to the screen, you can delete all of the duplicated code in test_nestedLoops

zkincaid:
You should also revise the regression tests as we discussed on Friday.

PR #15, opened 2020-01-16
Almost commuting

ahumenberger:
As a first step I implemented the naive abstraction algorithm which explores all partitions of the LTS (see PhasedSegmentation.make_naive).

Since I'm not familiar with best practices in OCaml, I don't know if it makes sense to wrap the functions in `almostCommuting.mli` in modules, or if this is an overkill.

### DIFF
--- a/srk/_oasis
+++ b/srk/_oasis
@@ -12,7 +12,7 @@ Library srk
   Path:             src
   FindlibName:      srk
   Pack:             true
-  Modules:  SrkUtil, Log, DisjointSet, Memo, FeatureTree, QQ, ZZ, Syntax, Interval, Smt, SrkZ3, Ring, Linear, Polynomial, ExpPolynomial, Interpretation, SrkApron, Polyhedron, SrkSimplify, Abstract, Nonlinear, CoordinateSystem, Wedge, Vas, Vass, Quantifier, Iteration, SolvablePolynomial, LinearSemiautomaton, Transition, BigO, Pathexpr, WeightedGraph, TransitionSystem
+  Modules:  SrkUtil, Log, DisjointSet, Memo, FeatureTree, QQ, ZZ, Syntax, Interval, Smt, SrkZ3, Ring, Linear, Polynomial, ExpPolynomial, Interpretation, SrkApron, Polyhedron, SrkSimplify, Abstract, Nonlinear, CoordinateSystem, Wedge, Vas, Vass, Quantifier, Iteration, SolvablePolynomial, LinearSemiautomaton, AlmostCommuting, Transition, BigO, Pathexpr, WeightedGraph, TransitionSystem
   BuildDepends:     batteries, ppx_deriving, ppx_deriving.show, ppx_deriving.ord, ppx_deriving.eq, gmp, camlidl, apron, Z3, ocrs, ntl, ocamlgraph
   XMETADescription: Symbolic Reasoning Kit
 
--- a/srk/src/almostCommuting.ml
+++ b/srk/src/almostCommuting.ml
@@ -84,7 +84,16 @@ module PhasedSegment = struct
         Array.map
           (fun (k, m) ->
             if k == Reset then
-              max_lds ~zero_rows:true mS (QQMatrix.mul mT m)
+              let mC, mD = pushout (QQMatrix.mul mT m) mS in
+              (* pushout TA S returns C,D such that CTA = DS. We then take the
+                 rowspace intersection of the matrices CT for all A, and iterate
+                 with T' = \bigcap CT. A fixpoint should be reached after two
+                 iterations, i.e. we have rsp(T') = rsp(T). If the fixpoint is
+                 reached, then there exists a C' s.t. C'CT = T. We use C' to
+                 perform a change of basis in order to get TA = C'DS. *)
+              match divide_right mT (QQMatrix.mul mC mT) with
+              | Some mC' -> mC, (QQMatrix.mul mC' mD)
+              | None     -> mC, mD
             else if k == Commute then
               max_lds mT (QQMatrix.mul mT m)
             else
--- a/srk/src/almostCommuting.ml
+++ b/srk/src/almostCommuting.ml
@@ -1,0 +1,21 @@
+open Linear
+open Printf
+
+type kind = Commute | Reset | Ignore
+
+type phased_segment = {
+   sim1 : QQMatrix.t;
+   sim2 : QQMatrix.t;
+   phase1 : QQMatrix.t array;
+   phase2 : (kind * QQMatrix.t) array
+}
+
+type phased_segmentation = phased_segment list
+
+
+let commuting_space mA mB =
+  let dims = SrkUtil.Int.Set.elements (QQMatrix.row_set mA) in
+  let mAB = QQMatrix.mul mA mB in
+  let mBA = QQMatrix.mul mB mA in
+  let mC = QQMatrix.add mAB (QQMatrix.scalar_mul (QQ.negate QQ.one) mBA) in
+  nullspace mC dims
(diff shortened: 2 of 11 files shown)


srk/_oasis
```diff
@@ -12,7 +12,7 @@ Library srk
   Path:             src
   FindlibName:      srk
   Pack:             true
-  Modules:  SrkUtil, Log, DisjointSet, Memo, FeatureTree, QQ, ZZ, Syntax, Interval, Smt, SrkZ3, Ring, Linear, Polynomial, ExpPolynomial, Interpretation, SrkApron, Polyhedron, SrkSimplify, Abstract, Nonlinear, CoordinateSystem, Wedge, Vas, Vass, Quantifier, Iteration, SolvablePolynomial, LinearSemiautomaton, Transition, BigO, Pathexpr, WeightedGraph, TransitionSystem
+  Modules:  SrkUtil, Log, DisjointSet, Memo, FeatureTree, QQ, ZZ, Syntax, Interval, Smt, SrkZ3, Ring, Linear, Polynomial, ExpPolynomial, Interpretation, SrkApron, Polyhedron, SrkSimplify, Abstract, Nonlinear, CoordinateSystem, Wedge, Vas, Vass, Quantifier, Iteration, SolvablePolynomial, LinearSemiautomaton, AlmostCommuting, Transition, BigO, Pathexpr, WeightedGraph, TransitionSystem
```
zkincaid, line 15: Can you make this change to ./_oasis as well?

srk/src/linear.ml
```diff
@@ -382,7 +382,7 @@ let max_rowspace_projection a b =
       | None -> ()));
   !c
 
-let max_lds mA mB =
+let max_lds ?zero_rows:(zero_rows=false) mA mB =
```
zkincaid: Why do we want to allow zero rows in the simulation?

srk/src/almostCommuting.ml
```diff
@@ -0,0 +1,146 @@
...
+      (fun vU s -> VS.sum vU (VS.of_matrix s.sim2))
+      VS.empty
+      segmentation
+
+  let almost_commutes segmentation =
+    let vU = almost_commuting_space segmentation in
+    let dim = VS.dimension vU in
+    VS.equal vU (VS.standard_basis dim)
+
+  let rec best_almost_commuting matrices =
```
zkincaid: 1. Does this need to be recursive?
2. I think this should return an LTS + simulation rather than a phased segmentation of that LTS
3. divide_right (mC * m) mC will give you a the correct (square) matrix:
(divide_right (mC * m) mC) * mC = mC * m

srk/src/almostCommuting.ml
```diff
@@ -0,0 +1,146 @@
...
+    else
+      let _, mA = Array.get pairs 0 in
+      let dims = SrkUtil.Int.Set.elements (QQMatrix.column_set mA) in
+      let mS, phase1 = commuting_segment (iter_all pairs) dims in
+      let mT, _ = commuting_segment (iter_commute pairs) dims in
+      let rec fix mT =
+        let ls_maxlds = Array.map
+          (fun (k, m) ->
+            if k == Reset then
+              max_lds ~zero_rows:true mS (QQMatrix.mul mT m)
```
zkincaid: This doesn't look right to me -- in general, if k = reset, then we should expect the associated matrix to be m x n, where m is the dimension of the almost commuting space and n is the dimension of the commuting space (max_lds will always produce a square matrix + a simulation).

ahumenberger: Assuming we call `max_lds mS mC` where S is a n x m matrix and C an o x m matrix with n <= o <=m. Then this gives me a simulation T and a matrix C' s.t. TC = C'TS.
If I understood the code of `max_lds` correctly, then `max_lds` implicitly treats its input matrices as square matrices, and then removes coordinates which correspond to zero rows in T*S. 
For instance the following
```ocaml
let mS = mk_matrix [[0; 1; 0]; 
                    [0; 0; 1]] 
in
let mC = mk_matrix [[0; 0; 2];
                    [0; 1; 2];
                    [0; 0; 1]]
in
max_lds mS mC
```
produces a simulation `[[1; 0]; [0; 1]]` and a square matrix `[[0; 2]; [1; 2]]`.
However, if we do not remove the coordinates corresponding to zero rows in TS (i.e. `max_lds ~zero_rows:true mS mC`), we get a simulation `[[1; 0; 0]; [0; 1; 0]; [0; 0; 1]` and a matrix C' (with implicit zeroes in the third column)
```ocaml
[[0; 2]; 
 [1; 2]; 
 [0; 1]]
```
We then have C'S = C.


Another example:
```ocaml
let mS = mk_matrix [[0; 0; 1]] in
let mC = mk_matrix [[0; 0; 2];
                    [0; 0; 1]]
in
max_lds ~zero_rows:true mS mC
```
gives a simulation `[1; 0]; [0; 1]` and a matrix `[[2]; [1]]`. `~zero_rows:false` would give a simulation `[1]` and a matrix `[2]`.

srk/_oasis
ahumenberger: Done.

srk/src/almostCommuting.ml
zkincaid: In
```ocaml
let mS = mk_matrix [[0; 1; 0]; 
                    [0; 0; 1]] 
in
let mC = mk_matrix [[0; 0; 2];
                    [0; 1; 2];
                    [0; 0; 1]]
in
max_lds mS mC
```
we have a transition system Sx' = Cx, corresponding to
```
y' = 2z
z' = y + 2z
0 = z
```
That is, the transition system is partial (defined only on the subspace where z = 0).  You shouldn't encounter such partial transition systems, because you are given as input a total transition system.
More generally, a zero row indicates non-existence of a best abstraction -- choosing the drop zero rows is an arbitrary choice (that gives us an abstraction that's defined anywhere) that we need to make when computing an abstraction of a non-total transition relation, but if best abstractions don't exist somewhere else in the pipeline then something is going wrong.

ahumenberger: In the case where `k==Reset` I'm not considering a transition system. Instead mS is the simulation where all matrices commute. and mC corresponds to T*A where T is the simulation where the nonreset matrices commute and A is a matrix in the reset part. I'm using `max_lds` to find a T' and a A' s.t. T'TA = A'T'S. If T' is not the idendity, then I have to refine T with T'T, otherwise A'T' would be \rho(A) as in the definition of phased segments.

zkincaid: I see... I think there's no need for a fixpoint computation here: you can find the determined functions of the Resets, and then find the commuting space of the Commutes, and then you're done.

ahumenberger: Right. I forgot to add a comment saying that the fixpoint should be reached after at most two iterations. Anyway, I changed the code to get rid of the unnecessary recursion.

srk/src/almostCommuting.ml
ahumenberger: I changed it according to your suggestions.

srk/src/linear.ml
ahumenberger: Should I create a separate function for the case `zero_rows=true` for the sake of having a cleaner interface? Right now, I'm somehow misusing max_lds for computing the determined functions for the resets.

zkincaid: If you just want the determined functions, you can use the pushout: supposing that `A` is an m-by-n linear map and `X` is a matrix whose rows form a basis for some subspace of Q^n (say of dimension k, so X is a k-by-n matrix), then `pushout A X` returns a pair `(D,C)` which is the "universal solution" to the equation `DA  = CX`.  This means that the rows of `D` form a basis for the space of vectors `d` such that `dA` belongs to the rowspace of X (`dA = cX` for some `c`).

srk/src/almostCommuting.ml
ahumenberger: Using `max_lds` here, does not work in general. I have to use `pushout`. Already changed.

srk/src/linear.ml
ahumenberger: I removed this optional argument again, as I'm now using `pushout` instead of `max_lds`.

PR #19, opened 2020-09-23
Linchc

JakeSilverman:
Logic for parsing integer CHCs and for reasoning about linear integer CHCs.


duet.opam
```diff
@@ -1,31 +1,27 @@
+# This file is generated by dune, edit dune-project instead
```
zkincaid: This file is not autogenerated.  These changed break the travis-ci build, and should be reverted.

srk/src/bigtop.ml
```diff
@@ -211,6 +214,24 @@ let spec_list = [
    ],
    " Generate a random formula");
 
+  ("-reachable-goal",
```
zkincaid: "-reachable-goal" is an unintuitive name.  "-chc"?

srk/src/bigtop.ml
```diff
@@ -211,6 +214,24 @@ let spec_list = [
...
+   Arg.String (fun file ->
+       let open Iteration in 
+       let chc = load_chc file in
+       let pd = 
+         (module Product(LinearRecurrenceInequation)(PolyhedronGuard) : PreDomain) 
+       in (*TODO: let user pick iter operation*)
+       let wg, init, goal = Chc.to_weighted_graph srk chc pd in
```
zkincaid: Don't use to_weighted_graph here.

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
+(* Constrainted horn clauses *)
```
zkincaid: Typo

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+  DynArray.add rel_ctx (name, typ);
+  DynArray.length rel_ctx - 1
+let type_of rel_ctx rel = snd (DynArray.get rel_ctx rel)
+let name_of rel_ctx rel = fst (DynArray.get rel_ctx rel)
+
+let rel_of_atom (relation, _) = relation
+let params_of_atom (_, params) = params
+
+let mk_rel_atom srk rel_ctx rel syms =
+  BatList.iter2
```
zkincaid: Can just use `(type_of rel_ctx rel) = (List.map (typ_symbol srk) syms)` to check equality.

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+             wg
+             rel
+             alg.one
+             goal_vert)
+        wg
+        chc.queries
+    in
+    wg
+
+  let has_reachable_goal srk chc pd =
```
zkincaid: Call this fn `check` for consistency with the rest of z3/srk.

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
+(* Constrainted horn clauses *)
```
zkincaid: Typo

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
...
+
+(** This function includes typechecking. We hide the type of relation atoms 
+ * to force the use to have their atom type checked *)
+val mk_rel_atom : 'a context -> relcontext -> relation -> symbol list -> 
+  relation_atom
+
+val rel_of_atom : relation_atom -> relation
+val params_of_atom : relation_atom -> symbol list
+
+val mk_hypo : relation_atom list -> 'a formula -> 'a hypothesis
```
zkincaid: delete

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
+(* Constrainted horn clauses *)
+open Syntax
+
+type relcontext
+type relation
+type relation_atom
+type 'a hypothesis = relation_atom list * 'a formula
```
zkincaid: delete

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
...
+type 'a rule = 'a hypothesis * relation_atom
```
zkincaid: delete

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
...
+  module Set : BatSet.S with type elt = relation
+  val compare : t -> t -> int
+  val pp : relcontext -> Format.formatter -> relation -> unit
+  val show : relcontext -> relation -> string
+end
+module Chc : sig
+  type 'a t = 'a chc
+  (** Create chc with no rules/queries *)
+  val create : 'a chc
+  val chc_of : 'a rule list -> query list -> 'a chc
```
zkincaid: delete

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
...
+  val add_rule : 'a chc -> 'a rule -> 'a chc
+  (** Adds query to chc and returns a fresh name for query *)
+  val add_query : 'a chc -> query -> 'a chc
+  val get_rules : 'a chc -> 'a rule list
+  val get_queries : 'a chc -> query list 
+  val get_relations_used : 'a chc -> Relation.Set.t
```
zkincaid: Why is this part of CHC rather than relcontext?

srk/src/syntax.mli
```diff
@@ -62,6 +62,8 @@ val int_of_symbol : symbol -> int
 
 val symbol_of_int : int -> symbol
 
+val dup_symbol : 'a context -> symbol -> symbol
```
zkincaid, line 67: +comment

srk/test/test_chc.ml
```diff
@@ -0,0 +1,192 @@
+open Srk
```
zkincaid, line 1: Can make tests more succinct with Syntax.Infix

srk/src/chc.mli
```diff
@@ -0,0 +1,89 @@
...
+  val pp : 'a context -> relcontext -> Format.formatter -> 'a chc -> unit
+  val show : 'a context -> relcontext -> 'a chc -> string
+  (** [is_linear chc] returns true if chc is linear chc - that is, each
+   * hypothesis has at most one relation atom *)
+  val is_linear : 'a chc -> bool
+  (** [has_reachable_goal srk chc pd] returns unknown if a query relation can
+   * be reached in the chc where recursion over-approximated using the 
+   * star operator of the provided predomain [pd] and returns no otherwise.*)
+  val has_reachable_goal : 
+    'a context -> 'a chc -> (module Iteration.PreDomain) -> [> `No | `Unknown]
```
zkincaid: Type should allow "Yes", even if that isn't supported by the current back-end.

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+        BatArray.fold_lefti 
+          (fun map ind asym -> Symbol.Map.add asym (mk_const srk (b.(ind))) map)
+          Symbol.Map.empty
+          a
+      in
+      let is_one (_, _, phi) = Formula.equal phi (mk_true srk) in
+      let is_zero (_, _, phi) = Formula.equal phi (mk_false srk) in
+      let zero = emptyarr, emptyarr, mk_false srk in
+      let one = emptyarr, emptyarr, mk_true srk in
+      let add x y =
```
zkincaid, line 119: Looks like this assumes that pre and post are shared by x and y.  How is property maintained?  What is the purpose of the substitution in add?

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+          let (pre1, post1, phi1) = x in
+          let (pre2, post2, phi2) = y in
+          let pre' = Array.map (fun sym -> dup_symbol srk sym) pre1 in
+          let post' = Array.map (fun sym -> dup_symbol srk sym) post2 in 
+          let rhs_pre_postmap = 
+            map_union (mk_subst_map post2 post') (mk_subst_map pre2 post1)
+          in
+          let rhs_subst =
+            Memo.memo 
+              ~size:(Symbol.Set.cardinal (symbols phi2))
```
zkincaid: https://cseweb.ucsd.edu/~kube/cls/100/Lectures/lec16/lec16-8.html

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+              (fun sym -> 
+                 if Symbol.Map.mem sym rhs_pre_postmap then
+                   Symbol.Map.find sym rhs_pre_postmap
+                 else (
+                   mk_const 
+                     srk
+                     (mk_symbol
```
zkincaid: dup_symbol?

srk/src/chc.ml
```diff
@@ -0,0 +1,408 @@
...
+open Syntax
+module DynArray = BatDynArray
+
+type relation = int
+type relcontext = (string * typ list) DynArray.t
+type relation_atom = relation * symbol list
+type 'a hypothesis = relation_atom list * 'a formula
+type 'a rule = 'a hypothesis * relation_atom
+type query = relation
+type 'a chc = {rules : 'a rule list; queries : query list}
```
zkincaid: "rule" and "chc" are synonyms.

srk/src/chc.ml
JakeSilverman: We assume x and y have the same size pre/post but not that the pre/post are equal in x and y. The substitution ensures x and y share the same pre/post at the conclusion of add. Specifically, we introduce a new pre/post (pre' and post') and subst xpre -> pre', xpost -> post' in xphi and subst ypre -> pre', ypost -> post' in yphi.

srk/src/chc.ml
JakeSilverman: Is this true even when I know the exact number of buckets ahead of time (as I do in this case)? It seems like when you know exact number there is no reason to over-estimate (assuming the table does not change size until it needs to).

srk/src/chc.mli
JakeSilverman: This function is analogous to "symbols" in Syntax.mli. That is, it gives the relation symbols that are used in a particular fixpoint. Relations can exist in the relcontext and not be used in a fixpoint in the same way that symbols can exist in a srk context and not be used in an expression.

srk/src/chc.ml
zkincaid: You always want to over-estimate, unless you have a perfect hash function.  Even if you do have a perfect hash function, you may still want to over-estimate if you're using a hash table from a library, since that hash table will almost certainly resize the underlying table when it gets close to being full.

srk/src/chc.mli
zkincaid: Why do we want relcontext to be a separate entity (i.e., what motivates divergence from Z3's API)?

PR #21, opened 2021-02-16
termination analysis, LLRF, phase analysis, exp, and DTA

happypig95:


bench/README.md
```diff
@@ -75,3 +75,20 @@ task suite named `SUITE`.
 
 Benchmark tasks should be defined by [YAML](https://yaml.org/) files,
 formatted according to [benchexec specifications](https://github.com/sosy-lab/benchexec/blob/master/doc/task-definition-example.yml)
+
+Suite descriptions
```
zkincaid, line 79: Add description for termination-linear

bench/benchmark-defs/2ls.xml
```diff
@@ -0,0 +1,37 @@
...
+  </tasks>
+  <tasks name="bitprecise">
+    <includesfile>../tasks/bitprecise.set</includesfile>
+    <propertyfile>../tasks/properties/termination.prp</propertyfile>
+    <option name="--64"/>
+    <option name="--termination"/>
+  </tasks>
+  <tasks name="recursive">
+    <includesfile>../tasks/recursive.set</includesfile>
+    <propertyfile>../tasks/properties/termination.prp</propertyfile>
```
zkincaid, line 27: --64 and --termination needed

duet/cra.ml
```diff
@@ -556,6 +565,7 @@ module TSDisplay = ExtGraph.Display.MakeLabeled
       let pp formatter w = match w with
         | Weight w -> K.pp formatter w
         | Call (s,t) -> Format.fprintf formatter "call(%d, %d)" s t
+      (* let show = SrkUtil.mk_show pp *)
```
zkincaid: delete

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
     end
   | _ -> assert false
 
+module TermMemo = Memo.Make(struct
+                      type t = (Ctx.t,Syntax.typ_fo) Syntax.expr
+                      let hash = Syntax.Expr.hash
+                      let equal = Syntax.Expr.equal
+                    end)
+
+let purify_floor =
```
zkincaid: Why not purify_floor from SrkSimplify?

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+  in
+  mk_and srk [purify_floor (K.guard transition);
+              substitute_const srk subst formula]
+
+(* Attractor region analysis *)
+let formula_with_attractor_region tf =
+  logf "Starting attractor region analysis";
+  let formula = TF.formula tf in
+  let xp_leq_x_terms = BatList.fold_left (fun l (x, xp) -> 
+      let open Syntax in (mk_leq srk (mk_const srk xp) (mk_const srk x), mk_const srk xp, mk_const srk x) :: l) [] (TF.symbols tf)
```
zkincaid: open Syntax in scope of the whole function so you don't need to open it repeatedly

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+  in
+  (* TODO: add an incremental interface for optimize_box *)
+  let lower_bounds = BatList.map (fun (xp_leq_x_term, xp, x) ->
```
zkincaid: Style: this is structured as the composition of several maps (map f (map g ( ...))).  Better to compose the functions you're mapping (f , g, ...) to make it easier to read and avoid allocating the intermediate lists.

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+let formula_with_attractor_region tf =
+  logf "Starting attractor region analysis";
+  let formula = TF.formula tf in
+  let xp_leq_x_terms = BatList.fold_left (fun l (x, xp) -> 
+      let open Syntax in (mk_leq srk (mk_const srk xp) (mk_const srk x), mk_const srk xp, mk_const srk x) :: l) [] (TF.symbols tf)
+  in
+  (* TODO: add an incremental interface for optimize_box *)
+  let lower_bounds = BatList.map (fun (xp_leq_x_term, xp, x) -> 
+      match SrkZ3.optimize_box srk (Syntax.mk_and srk [formula; xp_leq_x_term]) [xp] with
+      | `Sat l ->  let h = BatList.hd l in (Interval.lower h, x)
```
zkincaid: style: collapse into the pattern "`Sat [ivl] -> Interval.lower ivl"

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+} 
+
+let omega_algebra =  function
+  | `Omega transition ->
+     (** over-approximate possibly non-terminating conditions for a transition *)
+     begin
+       let open Syntax in
+       let tf =
+         TF.map_formula
+           (fun phi -> purify_floor (Nonlinear.linearize srk phi))
```
zkincaid: Do we want to purify floor here?

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+            let result, _ = TLLRF.compute_swf srk tf in 
+             if result = TLLRF.ProvedToTerminate then
+              begin
+                logf "proved to terminate by LLRF";
+                [Syntax.mk_false srk], true
+              end
+             else
+              let pre =
+                let fresh_skolem =
+                  Memo.memo (fun sym ->
```
zkincaid: use Syntax.dup_symbol

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+            let mp =
+              Syntax.mk_not srk
+                (TerminationExp.mp (module Iteration.LinearRecurrenceInequation) srk tf)
+            in
+            let dta_entails_mp =
+              (* if DTA |= mp, DTA /\ MP simplifies to DTA *)
+              Syntax.mk_forall_consts
+                srk
+                (fun _ -> false)
+                (Syntax.mk_or srk [mk_not srk (mk_and srk dta); mp])
```
zkincaid: Syntax.mk_if

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+          else []
+        in
+        let exp =
+          if (not llrf_succ) && !termination_exp then
+            let mp =
+              Syntax.mk_not srk
+                (TerminationExp.mp (module Iteration.LinearRecurrenceInequation) srk tf)
+            in
+            let dta_entails_mp =
+              (* if DTA |= mp, DTA /\ MP simplifies to DTA *)
```
zkincaid: Why do we check  this?

duet/cra.ml
```diff
@@ -713,6 +723,312 @@ let analyze file =
...
+      | _ -> []
+    ) 
+      lower_bounds 
+      upper_bounds
+  in
+  let formula'' = Syntax.mk_and srk (formula :: BatList.flatten lb_x_ub) in
+  logf "Formula with attractor regions:\n%a\n" (Syntax.Formula.pp srk) formula'';
+  TF.map_formula (fun _ -> formula'') tf
+
+let omega_algebra_for_phase srk nonterm = WG.{
```
zkincaid: Seems like it should be somewhere in srk.  Not really just the omega algebra for phase -- this is the generic omega algebra construction for termination analysis

duet/dune
```diff
@@ -6,7 +6,7 @@
   (preprocess (pps ppx_deriving.ord ppx_deriving.eq ppx_deriving.show)))
 
 (ocamllex
-  (modules cbpLex))
```
zkincaid, line 9: Revert?

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+       in
+       Smt.Solver.add solver [mk_not srk cell_formula];
+       find_cells (cell_formula::cells)
+    | `Unsat -> cells
+    | `Unknown -> assert false (* to do *)
+  in
+  find_cells []
 
+  (* Each cell is i, (pos_pred_indices, neg_pred_indices), cell_formula *)
+let invariant_complex_partition srk candidates tf =
```
zkincaid: Refactor so we don't need both invariant_partition and invariant_complex_partition?  (Are these the same (as they appear), or is there some sense in which this is "complex"?)

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+        logf "transition formula UNSAT, no further cells";
+        cells
+      | `Unknown -> assert false (* to do *)
+    in
+    predicates, find_cells []
+  
+(*  Ranked cells is a map from int to (list of sets), where int is the number of 
+    positive predicates.
+*)
+let rank_cells cells =
```
zkincaid: Internal to build_graph?

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+              let indicator_name_px = "ind_2_for_pred_" ^ (string_of_int ind) in  
+              let indicator_symbol = mk_symbol srk ~name:indicator_name_px `TyBool in 
+              mk_not srk (mk_const srk indicator_symbol)
+            ) 
+        |> mk_and srk 
+      in 
+      let cell2_can_follow_cell1 = mk_and srk [cond_neg_clause; cond_pos_clause; result_neg_clause; result_pos_clause] in
+      List.exists (fun m -> Interpretation.evaluate_formula m cell2_can_follow_cell1) !models || 
+      begin
+        (* use the Solver.check to avoid pushing and popping, pass in a list of boolean vars *)
```
zkincaid: If you *don't* use the assumption interface, then you don't need to use indicator vars

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+  } 
+  in
+  let wg = ref (WG.empty algebra) in
+  (* node 0 as the virtual entry *)
+  wg := WG.add_vertex !wg 0;
+  (* virtual entry -> every cell with 0 indegree: weight 1
+      self-loop for every cell: weight tf /\ cell_formula
+      edge from cell A to cell B: weight 1
+   *)
+  let combine tf f = TF.make ~exists:(TF.exists tf) (mk_and srk [TF.formula tf; f]) (TF.symbols tf) in
```
zkincaid: TF.map (fun g -> mk_and srk [f; g]) tf

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+    (fun m i ((positive_preds, negative_preds), f) -> 
+      let num_pos_preds = BatSet.Int.cardinal positive_preds in
+      match BatMap.Int.find_opt num_pos_preds m with 
+      | Some l -> BatMap.Int.update num_pos_preds num_pos_preds ((i, (positive_preds, negative_preds), f)::l) m
+      | None -> BatMap.Int.add num_pos_preds [(i, (positive_preds, negative_preds), f)] m
+    )
+  ranked_cells
+  cells
+
+let build_graph_and_compute_mp srk tf inv_predicates omega_algebra ranked_cells num_cells =
```
zkincaid: Factor out the construction of the graph?

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+  in
+  let wg = ref (WG.empty algebra) in
+  (* node 0 as the virtual entry *)
+  wg := WG.add_vertex !wg 0;
+  (* virtual entry -> every cell with 0 indegree: weight 1
+      self-loop for every cell: weight tf /\ cell_formula
+      edge from cell A to cell B: weight 1
+   *)
+  let combine tf f = TF.make ~exists:(TF.exists tf) (mk_and srk [TF.formula tf; f]) (TF.symbols tf) in
+  let zero_indegree_vertices = ref (BatSet.Int.empty) in
```
zkincaid: Just compute at the end rather than trying to maintain this structure?

srk/src/iteration.ml
```diff
@@ -688,8 +689,323 @@ let invariant_transition_predicates srk tf predicates =
...
+        begin
+          BatList.iter (fun (j, (pos_preds_j, neg_preds_j), _) -> 
+            begin
+            logf "checking if cell %d could be followed by cell %d" i j;
+            if not (BatSet.Int.mem j (descendants.(i))) then
+              logf "j is not one of i's descendants, check if i -> j";
+              if can_follow (pos_preds_i, neg_preds_i) (pos_preds_j, neg_preds_j) solver models then 
+                begin
+                  logf "cell %d could be followed by cell %d" i j;
+                  (* Cell # i is the (i+1)-th vertex in the graph *)
```
zkincaid: Why?

srk/src/iteration.mli
```diff
@@ -71,3 +71,22 @@ module ProductWedge (A : PreDomainWedge) (B : PreDomainWedge) : PreDomainWedge
...
+   in the sense that when a computation leaves a cell it may never return.
+   This function takes a set of candidate transition predicates as input,
+   determines which are invariant, and returns the non-empty cells of the
+   partition. *)
+val invariant_partition : 'a context ->
+                          ('a formula) list ->
+                          'a TransitionFormula.t ->
+                          ('a formula) list
+
+val compute_mp_with_phase_DAG : 'a context ->
```
zkincaid: Description?

srk/src/termination.ml
```diff
@@ -0,0 +1,292 @@
+(* Linear ranking function synthesis, subsumed by LLRF synthesis. *)
```
zkincaid: Is this file used anywhere?  If not, delete.

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+module Vec = Linear.QQVector
+module Mat = Linear.QQMatrix
+module TF = TransitionFormula
+
+include Log.Make(struct let name = "TerminationDTA" end)
+
+(** Create symbols that stands for some linear terms and their defining
+    equalities.
+*)
+let build_symbols_for_inv_terms srk inv_terms =
```
zkincaid: In what sense are these "invariant terms"?

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+           mk_symbol srk ~name:name_str (expr_typ srk term)
+         in
+         let const_expr = mk_const srk symbol in
+         let equality = mk_eq srk term const_expr in
+         (symbol :: l_symbols, equality :: l_equalities, Symbol.Set.add symbol s)
+      )
+      inv_terms
+      ([], [], Symbol.Set.empty )
+  in
+  (l_symbols, l_equalities, symbol_set)
```
zkincaid: Style: don't name a value and then return the same -- just return the value.

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  let module PLM = Lts.PartialLinearMap in
+  let omega_domain = snd (PLM.iteration_sequence best_dlts.dlts) in
+  let rep = PLM.map (PLM.make (PLM.map best_dlts.dlts) omega_domain) in
+  rep
+
+
+(** Compute the exponential polynomial that represents the transitive
+    closure of a DLTS.
+*)
+let compute_exp_polynomial ?(square=false) m =
```
zkincaid: Doesn't make sense for square to be a parameter -- just call compute_exp_polynomial on m or (m * m) as appropriate.
This function doesn't really do anything -- delete and use ExpPolynomial.exponentiate_rational instead

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  let best_dlts_tr_matrix = if square then Linear.QQMatrix.mul m m else m in
+  logf "getting squared is: %b" square;
+  logf "matrix to be exponentiated: %a" (Linear.QQMatrix.pp) best_dlts_tr_matrix;
+  let exp_poly_mat = if Linear.QQMatrix.zero == m then ExpPolynomial.Matrix.zero else
```
zkincaid: Not necessary

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+(** This data structure is a map with key type (QQ.t, int) and value type QQVector.
+    Ordering on the keys is the lexicographic ordering, and value types
+    form an Abelian group.
+    This is used to store a term in an exponential polynomial. 
+    For example, a term could look like 2^k * (k^3) * (2x + y + z) where
+    k is the symbolic loop counter, and x, y, z are the initial values of
+    variables x, y, z.
+*)
+module BaseDegPairMap = struct
+  type pair = QQ.t * int [@@deriving ord]
```
zkincaid: Implement compare so that we don't depend upon [@@deriving ord] to derive lex order

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+   match ExpPolynomial.exponentiate_rational best_dlts_tr_matrix with
+    | None -> failwith "best rational DLTS abstraction produced a matrix with not all rational eigenvalues!"
+    | Some mat -> mat
+  in
+  logf "Printing exp poly:\n%a\n" ExpPolynomial.Matrix.pp exp_poly_mat;
+  exp_poly_mat
+
+type ineq_type = Lt0 | Eq0 | Leq0
+
+(** This data structure is a map with key type (QQ.t, int) and value type QQVector.
```
zkincaid: I can read this off of the line "module E = SparseMap.Make(struct type t = pair [@@deriving ord] end)(QQV)".  Comment should describe correspondence between the data structure and exponential polynomial terms

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+      p: the Abelian group map.
+      invariant_terms: these are the linear terms whose closed forms are computed
+          during the symbolic exponentiation of transition matrix. This is 
+          used to interpret the meaning of the vectors as values in the map.
+      lhs_const: the constant that appears in the closed form formula.
+      ineq_type: whether this inequality is <=, = or <.
+  *)
+  let rank srk p invariant_terms lhs_const ineq_type =
+    logf "entering rank";
+    let e = E.enum p in
```
zkincaid: Style: it's almost never desirable to name an enum since they are transient.

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    let dim_to_term = fun d -> if d < 0 then mk_one srk else BatList.nth invariant_terms d in
+    let dom_ranked_list =
```
zkincaid, line 81: BatList.of_enum (E.enum p) if you want a list -- but it looks like ultimately you're just folding over this list -- you can just fold over the enum instead.

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+        ) 
+        []
+        e
+    in
+    let conditions = 
+      logf "start printing conditions list and formula stems";
+      let conditions_list_final, formula_stem_list, encountered_const_order = 
+        BatList.fold_left
+          (fun (conditions_list, formula_stem, flag) (base, deg, qqt)  ->
+             if has_const_order base deg then
```
zkincaid: Do we need this special case?

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  (** Rank the items in the data structure according to dominance ordering and
+      produce a formula for sufficient condition for termination.
+      p: the Abelian group map.
+      invariant_terms: these are the linear terms whose closed forms are computed
+          during the symbolic exponentiation of transition matrix. This is 
+          used to interpret the meaning of the vectors as values in the map.
+      lhs_const: the constant that appears in the closed form formula.
+      ineq_type: whether this inequality is <=, = or <.
+  *)
+  let rank srk p invariant_terms lhs_const ineq_type =
```
zkincaid: don't need lhs_const -- can just subtract (p - lhs_const) so that the comparison is alway with zero.
I think the spec should be something like: compute a formula describing the initial states for which "p ineq_typ 0" is eventually invariant.

srk/src/terminationDTA.ml
zkincaid: What is flag for?

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  logf "matrix to be exponentiated: %a" (Linear.QQMatrix.pp) best_dlts_tr_matrix;
+  let exp_poly_mat = if Linear.QQMatrix.zero == m then ExpPolynomial.Matrix.zero else
+   match ExpPolynomial.exponentiate_rational best_dlts_tr_matrix with
+    | None -> failwith "best rational DLTS abstraction produced a matrix with not all rational eigenvalues!"
+    | Some mat -> mat
+  in
+  logf "Printing exp poly:\n%a\n" ExpPolynomial.Matrix.pp exp_poly_mat;
+  exp_poly_mat
+
+type ineq_type = Lt0 | Eq0 | Leq0
```
zkincaid, line 36: Can eliminate the Lt0 case, since (for integers) x < y iff x + 1 <= y

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+      ([1, 0, 2], 1)
+    *)
+let get_coeff_vec_wrt_symbol_list expression list_symbols = 
+  let vec = 
+    BatList.fold_lefti
+      (fun existing_coeffs i symbol -> 
+        try
+          let coeff = Vec.coeff (Linear.dim_of_sym symbol) expression in
+          Vec.add_term coeff i existing_coeffs 
+        with Not_found -> Vec.add_term QQ.zero i existing_coeffs
```
zkincaid: What throws Not_found?

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    Linear.term_of_vec srk (fun i -> simulation.(i)) row
+    |> SrkSimplify.simplify_term srk)
+  |> BatArray.of_enum
+
+(** For a linear term, get its vector representation w.r.t. a list of symbols. 
+    For example, x + 2z + 1 w.r.t. [x, y, z] is represented by a tuple 
+    containing a vector and the constant term:
+      ([1, 0, 2], 1)
+    *)
+let get_coeff_vec_wrt_symbol_list expression list_symbols =
```
zkincaid: This looks like a change of basis: list_symbols is a new basis, expression is a vector in the standard basis (using the default correspondence between symbols and ints)

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+        with Not_found -> Vec.add_term QQ.zero i existing_coeffs 
+      )
+      Vec.zero
+      list_symbols
+  in
+  let const = Vec.coeff CoordinateSystem.const_id expression in
+  (vec, const)
+
+(* Scale a simulation such that it maps integer states to integer states *)
+let scaling_simulation srk best_DLTS_abstraction =
```
zkincaid: "scale_simulation"

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+      list_symbols
+  in
+  let const = Vec.coeff CoordinateSystem.const_id expression in
+  (vec, const)
+
+(* Scale a simulation such that it maps integer states to integer states *)
+let scaling_simulation srk best_DLTS_abstraction = 
+  let sim = best_DLTS_abstraction.simulation in 
+  let lcm_of_denominators = BatArray.fold_left (fun d term -> 
+    let coeff_vec = Linear.linterm_of srk term in
```
zkincaid: Lot's of unnecessary names (including one for an enum).
BatEnum.fold (fun dd (coeff, _) -> ZZ.lcm dd (QQ.denominator coeff)) ZZ.one (Linear.QQVector.enum coeff_vec)

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+let integer_spectrum_abstraction srk tr_matrix simulation = 
+let open Linear in
+let dims =
+  SrkUtil.Int.Set.union
+    (QQMatrix.row_set tr_matrix)
+    (QQMatrix.column_set tr_matrix)
+  |> SrkUtil.Int.Set.elements
+in
+let rsd = Linear.rational_spectral_decomposition tr_matrix dims in
+let int_eig_vecs, non_int_eig_vecs =
```
zkincaid: BatList.partition

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    Linear.QQVectorSpace.simplify
+      (Linear.QQVectorSpace.of_matrix (QQMatrix.mul int_eigenspace simulation_mat))
+    |> QQMatrix.of_rows
+  in
+  match Linear.divide_right int_eigenspace_sim simulation_mat with
+  | Some mS' -> mS'   (* simpl(ES) = mS' * mS*)
+  | None -> assert false
+in
+let dom_constraints_lhs = compose_simulation srk (QQMatrix.of_rows non_int_eig_vecs) simulation in
+let constraints = BatArray.fold_left (fun f term -> mk_and srk [f; mk_eq srk (mk_zero srk) term]) (mk_true srk) dom_constraints_lhs in
```
zkincaid: mk_and srk (List.map (mk_eq srk (mk_zero srk)) (BatArray.list_of dom_constraints_lhs))

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  let seq_of_true srk =
+    Periodic.make [mk_true srk]
+
+  let seq_of_false srk =
+    Periodic.make [mk_false srk]
+    
+  let seq_and srk x y = 
+    Periodic.map2 (fun a b -> mk_and srk [a; b]) x y
+
+  let seq_and srk xs =
```
zkincaid, line 252: Periodic.mapn (mk_and srk) xs

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    Periodic.map2 (fun a b -> mk_add srk [a; b]) x y
+
+  let seq_mul_symbol srk x symbol =
+    Periodic.map (fun f -> mk_mul srk [f; mk_const srk symbol]) x
+
+  let seq_not srk x = 
+    Periodic.map (fun f -> mk_not srk f) x
+
+  (* calculate the termination condition induced by characteristic sequence x *)
+  let seq_conditions srk x =
```
zkincaid: The "not" is unexpected for "seq_conditions".  I'd remove this function, actually -- it doesn't do much, and reading the definition in the context where it is used is probably easier than seeing an occurrence of seq_conditions

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    for i = 0 to (modulo - 1) do
+      let p_at_i = Polynomial.QQX.term_of srk (mk_int srk i) poly in
+      let p_at_i_mod = mk_mod srk p_at_i (mk_int srk modulo) in
+      seq := p_at_i_mod :: !seq 
+    done;
+    Periodic.make (BatList.rev !seq)
+
+  (* seq_of_polynomial srk m p b returns characteristic sequence of b^k p(k) mod m, 
+  k = 0, 1, 2, ... *)
+  let seq_of_single_base_exp_polynomial srk modulo poly base =
```
zkincaid: This evaluates to a sequence of ints -- why is does it return a term?

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  (* seq_of_exp m t returns the characteristic sequence of t^k mod m, 
+  k = 0, 1, 2, ... *)
+  let seq_of_exp modulus lambda = 
+    UltimatelyPeriodic.unfold (fun power -> (power * lambda) mod modulus) 1 
+    |> periodic_approx
+
+
+  (* seq_of_polynomial srk m p returns characteristic sequence of p(k) mod m, 
+  k = 0, 1, 2, ... *)
+  let seq_of_polynomial srk modulo poly =
```
zkincaid: This evaluates to a sequences of ints -- why does it return a term?
You can use Polynomial.QQX.eval to evaluate.
Can write more succinctly using
(0 -- (modulus - 1))
/@ (fun i -> match QQ.to_int (Polynomial.QQX.eval poly (QQ.of_int i)) with
| Some result -> result mod modulus
| None -> assert false)
|> BatList.of_enum
|> Periodic.make

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+            srk 
+            (mk_mul srk [ (mk_int srk n); term]) 
+            (mk_int srk modulo)
+      ) 
+    seq_of_exp
+    seq_of_poly
+
+  (* characteristic sequence of an exponential polynomial modulo some number *)
+  let seq_of_exp_polynomial srk modulo exppoly =
+    let e = ExpPolynomial.enum exppoly in
```
zkincaid: named enum

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+    seq_of_poly
+
+  (* characteristic sequence of an exponential polynomial modulo some number *)
+  let seq_of_exp_polynomial srk modulo exppoly =
+    let e = ExpPolynomial.enum exppoly in 
+    BatEnum.fold 
+      (fun existing_seq (poly, base) -> 
+        let b = match QQ.to_int base with 
+          Some i -> i 
+          | None -> failwith "Did not expect non-integer bases here"
```
zkincaid: to_int succeeds when base is a *63-bit* integer

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+        in
+        let current_seq = seq_of_single_base_exp_polynomial srk modulo poly b in 
+          seq_add srk existing_seq current_seq
+      )
+    (Periodic.make [mk_zero srk])
+    e 
+          
+  (* Compute characteristic sequence of atoms with form p < 0, p <= 0, or p = 0. *)
+  let generate_xseq srk lhs ineq_type exp_poly abstraction =
```
zkincaid: Difficult to match spec to the list of parameters

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+      Vec.pp lt_vec 
+      QQ.pp lhs_const;
+    let lt_vec_exppoly = ExpPolynomial.Vector.of_qqvector lt_vec in
+    let closed_form_vec = ExpPolynomial.Matrix.vector_left_mul lt_vec_exppoly exp_poly in
+    let mat = ExpPolynomial.Matrix.of_rows [closed_form_vec] in
+    logf "\nClosed form matrix is: %a, with %d rows and %d cols" 
+    ExpPolynomial.Matrix.pp mat (ExpPolynomial.Matrix.nb_rows mat) (ExpPolynomial.Matrix.nb_columns mat);
+    let mat_entries = ExpPolynomial.Matrix.entries mat in
+    (* check if the dynamic matrix has negative eigenvalues *)
+    let has_negative_base = BatEnum.exists
```
zkincaid, line 341: Logs seem unncessary.  Remove named enums

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,599 @@
...
+  (* Compute characteristic sequence of atoms with form p < 0, p <= 0, or p = 0. *)
+  let generate_xseq srk lhs ineq_type exp_poly abstraction = 
+    let (invariant_symbols, invariant_terms, best_DLTS_abstraction) = abstraction in
+    let lt_vec, lhs_const = get_coeff_vec_wrt_symbol_list lhs invariant_symbols in
+    logf "\nlt_vec is: %a, LHS constant is: %a" 
+      Vec.pp lt_vec 
+      QQ.pp lhs_const;
+    let lt_vec_exppoly = ExpPolynomial.Vector.of_qqvector lt_vec in
+    let closed_form_vec = ExpPolynomial.Matrix.vector_left_mul lt_vec_exppoly exp_poly in
+    let mat = ExpPolynomial.Matrix.of_rows [closed_form_vec] in
```
zkincaid: Cloed form "matrix" is just a vector -- why put it into a matrix?

srk/test/test_termination.ml
```diff
@@ -0,0 +1,63 @@
...
+          Infix.((!(x = y)
+                  && ((x' = x + (int 1) && y' = y)
+                      || (x' = x && y' = y - (int 1)))))
+        in
+        assert_equiv_quantified_formula
+          Infix.(x <= y)
+          (mp_exp tr_symbols tr)
+      );
+
+      "exp_seq1" >:: (fun () ->
```
zkincaid: Can delete these three -- repeats of tests in test_sequence

srk/test/test_termination.ml
```diff
@@ -0,0 +1,63 @@
...
+  TerminationExp.mp
+    (module Iteration.LinearRecurrenceInequation)
+    srk
+    (TransitionFormula.make
+       ~exists:(fun sym -> List.mem sym all_sym)
+       phi
+       tr_symbols)
+
+                 
+let suite = "Termination" >::: [
```
zkincaid, line 68: Only mp_exp is tested -- need more tests!

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
+open Syntax
+module Vec = Linear.QQVector
+module Mat = Linear.QQMatrix
+module TF = TransitionFormula
+
+include Log.Make(struct let name = "TerminationLLRF" end)
+
+type analysis_result = ProvedToTerminate | Unknown
```
zkincaid: isomorphic to bool -- just return true or false instead.

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+*)
+let get_polyhedron_of_formula srk f cs =
+  let f = match Formula.destruct srk f with
+    | `And xs -> xs
+    | `Tru -> []
+    | `Atom _ -> [ f ]
+    | _ -> failwith "formula is not convex polyhedron"
+  in
+  Polyhedron.of_implicant ~admit:true cs f
+  |> Polyhedron.enum
```
zkincaid: Polyhedron.enum does not exist anymore -- should be Polyhedron.enum_constraints

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+             ineqs
+         in
+         let coeff_sym = List.nth coeff_x_list i in
+         let equ_for_this_var = mk_eq srk (mk_const srk coeff_sym) full_rhs in
+         mk_and srk [ equ_for_this_var ; formulas]
+      )
+      f_with_non_neg_lambdas
+      dx_list
+  in
+  logf "\nformula of lambdas on non-inc:\n%s\n\n" (Formula.show srk answer);
```
zkincaid: Don't use show to pretty-print -- use pp.  show doesn't respect formatting constraints of surrounding context and it slows things done even when we don't end up printing to the terminal

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+    according to Farkas' lemma, every function of the form 
+    a(x - x') + b(y - y') >= 0 must be non-negative combinations of the 
+    half-spaces that consitute the boundaries of the polyhedron.
+    cs: coordinate system
+    ineqs: inequalities as list of vectors
+    dx_list: list of delta symbols defined using dx = x' - x for every x
+    coeff_x_list: a list of symbolic coefficients in the final linear terms
+    coeff_x_set: set of symbolic coefficients
+*)
+let build_system_of_ineq_for_non_inc srk cs ineqs dx_list coeff_x_list coeff_x_set =
```
zkincaid: Lots of parameters here, and I don't know how to differentiate this spec from the one for compute_non_inc_cone.  Seems that they should be combined and simplified.

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+    linear terms over pre-transition variables that are guaranteed to be 
+    bounded from below.
+    cs: coordinate system
+    ineqs: inequalities as list of vectors
+    x_list: list of variable symbols
+    coeff_x_list: a list of symbolic coefficients in the final linear terms
+    coeff_x_set: set of symbolic coefficients
+    TODO: extract common logic into dual cone computation into polyhedron.ml
+*)
+let build_system_of_ineq_for_lb_terms srk cs ineqs x_list coeff_x_list coeff_x_set =
```
zkincaid: Combine with compute_lower_bound_term_cone and simplify

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+  let ineqs = get_polyhedron_of_formula srk formula_of_lbx cs in
+  let non_inc_cone = build_system_of_ineq_for_lb_terms srk cs ineqs x_list coeff_x_list coeff_x_set in
+  non_inc_cone
+
+(** Compute quasi-ranking functions for formula f at a certain depth.
+    This is done by intersecting the two cones at each depth, and then constraining
+    the formula by dictating that the term for generators of the
+    intersection has to stay the same, then recurse to the next depth with the 
+    residual formula.
+    depth: we are currently synthesizing the depth-th component of a lexicographic ranking function
```
zkincaid: necessary?

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+    qrfs: quasi ranking functions
+    x_list, xp_list, x_set, xp_set: list/set of pre- and post-transition symbols
+    dx_set: set of delta symbols
+    x_to_dx, dx_to_x: get the corresponding delta symbol for a variable symbol and vice versa
+    coeff_x_list: list containing symbolic coefficients of each variable in the final term
+    coeff_x_set: set of symbolic coefficients for easy look up
+*) 
+let rec find_quasi_rf depth srk f qrfs x_list xp_list dx_list x_set xp_set dx_set x_to_dx dx_to_x coeff_x_list coeff_x_set =
```
zkincaid: Lots of parameters -- simplify

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+    that each dx = x' - x.
+*)
+let add_diff_terms_to_formula srk f x_xp =
+  List.fold_right
+    (fun (x, xp) (f, dx_list, dx_sym_set, x_to_dx, dx_to_x) -> 
+       let dname = String.concat "" ["d_"; show_symbol srk x] in
+       let cx = mk_const srk x in
+       let cxp = mk_const srk xp in 
+       let diff = mk_sub srk cxp cx in
+       let dx_sym = mk_symbol srk ~name:dname `TyInt in
```
zkincaid: should have same type as x -- not necessarily int.

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,425 @@
...
+
+(* let simplify_residual_formula srk x_set xp_set residual_formula = 
+  let polka = Polka.manager_alloc_strict () in
+  let exists = fun x -> (Symbol.Set.mem x x_set || Symbol.Set.mem x xp_set) in
+  Abstract.abstract ~exists:exists srk polka residual_formula
+  |> SrkApron.formula_of_property *)
+(** The actual swf operator only has true or false as outcomes, corresponding to
+    able to prove or unable to prove results given here.
+ *)
+let compute_swf srk tf =
```
zkincaid: Rename.  The formula that's returned is not a sufficient well-foundedness condition (it's a residual), and we should try our best to match the vocabulary in the paper.

srk/src/iteration.ml
happypig95: The 0-th vertex is occupied by the virtual start vertex that we added first.

srk/src/iteration.ml
happypig95: This large function is mostly for the construction of graph. The mp computation takes < 10 lines.

duet/cra.ml
happypig95: I think we are not actually running the attractor region analysis now. Maybe this is of lower priority?

duet/cra.ml
happypig95: I remembered that quantifier elimination in terminationExp was giving errors if we don't purify floor here. Maybe that has changed as well but I'm not sure.

srk/src/iteration.ml
happypig95: For larger graph we need to traverse each vertex and edge again to maintain this. And it is hard to traverse the graph to find incoming edges.

srk/src/terminationLLRF.ml
happypig95: I thought the current version is a bit more intuitive to understand.

srk/src/terminationDTA.ml
happypig95: We haven't implemented vector times matrix yet for exppolynomials, this is just a workaround.

srk/src/terminationDTA.ml
happypig95: If the symbol does not appear in the expression, Vec.coeff throws this.

srk/src/terminationDTA.ml
happypig95: Do we want to implement this as a change of basis? This looks like overkill to me.

srk/src/terminationDTA.ml
happypig95: Does this mean that non-strict inequalities do not occur? Or that I need to handle them but as Eq0?

srk/src/terminationDTA.ml
happypig95: This is used to deal with one scenario where the LHS has a first-order symbolic constant and the equation to be analyzed is like x0 + 5 = 0. I'm thinking if this can be simplified if I managed to subtract the constant from LHS

srk/src/iteration.ml
zkincaid: Could just use -1 is the virtual start vertex

srk/src/iteration.ml
zkincaid: Can't you just  use WG.fold_vertex to traverse vertices, and (WG.U.in_degree (WG.forget_weights wg) v) to check which have in-degree 0?

srk/src/terminationDTA.ml
zkincaid: Vec.coeff never throws Not_found -- conceptually, every dimension has a coefficient, but only finitely many dimensions have a non-zero coefficient.

srk/src/terminationDTA.ml
zkincaid: I suggest that you rename the function "change_basis" (or similar) to reflect the fact that this is what is happening -- you don't need to change the code.

srk/src/terminationDTA.ml
```diff
@@ -0,0 +1,586 @@
...
+  k = 0, 1, 2, ... *)
+  let seq_of_polynomial modulus poly = 
+    let lcm_of_denoms =
+      BatEnum.fold (fun current_lcm (coeff, _) -> ZZ.lcm current_lcm (QQ.denominator coeff)) 
+      ZZ.one
+      (Polynomial.QQX.enum poly)
+    in
+    let poly = Polynomial.QQX.scalar_mul (QQ.of_zz lcm_of_denoms) poly in
+    let m = modulus * (BatOption.get (ZZ.to_int lcm_of_denoms)) in
+    (0 -- (modulus - 1))
```
zkincaid, line 290: Should be m?  (Style: I would do "let modulus = modulus * ... in ... " to make sure I don't re-use the (incorrect) modulus).

srk/src/terminationDTA.ml
zkincaid: I don't see where you use this

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,194 @@
...
+    (* Origin must belong to the lower-bounded terms cone *)
+    BatEnum.push lb_cone_generators (`Vertex, Linear.QQVector.zero);
+    let generators_list = BatList.of_enum lb_cone_generators in
+    let lb_cone = Polyhedron.of_generators dual_space_dim (BatList.enum generators_list) in
+    lb_cone
+  in
+
+  logf "calculating meet of non-inc cone and lower-bounded cone";
+  let non_inc_cone = compute_non_inc_cone formula in
+  if Polyhedron.equal non_inc_cone Polyhedron.bottom then
```
zkincaid: Doesn't non_inc_cone always contain the origin?

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,194 @@
...
+                let qrf_unchange_constraints = BatList.map 
+                  (fun (gen_kind, vec) -> 
+                    match gen_kind with 
+                    | `Vertex -> mk_true srk
+                    | `Ray ->
+                      logf "looking at generator with vector: %a" Linear.QQVector.pp vec;
+                      let unprimed_exp = Linear.term_of_vec srk (fun d -> let x, _ = BatList.nth x_xp d in mk_const srk x) vec in
+                      let primed_exp = Linear.term_of_vec srk (fun d -> let _, xp = BatList.nth x_xp d in mk_const srk xp) vec in
+                      mk_eq srk unprimed_exp primed_exp
+                    | `Line ->
```
zkincaid: A line corresponds to a constant term (both vec and -vec are non-increasing, and therefore vec does not change).  Why do we equate vec' with 0?

srk/src/terminationLLRF.ml
```diff
@@ -0,0 +1,194 @@
...
+                let constrained_formula = mk_and srk (formula :: qrf_unchange_constraints) in
+                logf "Constrained formula for the next level is: %a" (Formula.pp srk) constrained_formula;
+                match Smt.equiv srk formula constrained_formula with
+                | `Unknown -> 
+                    logf ~attributes:[`Bold; `Red] "Cannot decide if there is any improvement at this level"; (false, formula)
+                | `Yes -> 
+                    logf ~attributes:[`Bold; `Red] "No improvement at this level, halt LLRF synthesis"; (false, formula)
+                | `No -> 
+                    logf ~attributes:[`Bold; `Green] "There is improvement at this level";
+                    match Smt.get_model srk constrained_formula with
```
zkincaid: We don't need this model

PR #20, opened 2020-12-31
Z3arr

JakeSilverman:
Still need to revise chc and arraycontent. Syntax and coordinate system need minor revisions as well. Arrays not working with simsat.

PR #23, opened 2021-04-02
added type info to substitute. fixed bug in substitute_sym

JakeSilverman:
Previous discussion:

https://github.com/zkincaid/duet/pull/20#discussion_r550509824
https://github.com/zkincaid/duet/pull/20#discussion_r550578419

A function may have more free vars than it does parameters in the case that the user wants an uncaptured var to appear in the resulting term. For example, consider if the user wants to replace the term "f(x)" with the term "f(x + var 0)". I added tests to cover previous bug and to show desire for having free vars beyond # of parameters. Note this is also in line with current specification in .mli file.

### DIFF
--- a/srk/src/syntax.ml
+++ b/srk/src/syntax.ml
@@ -362,11 +362,11 @@ let substitute srk subst sexpr =
     match label with
     | Exists (_, _) | Forall (_, _) ->
       go_children label (depth + 1) children
-    | Var (v, _) ->
+    | Var (v, typ) ->
       if v < depth then (* bound var *)
         sexpr
       else
-        decapture srk 0 depth (subst (v - depth))
+        decapture srk 0 depth (subst ((v - depth), typ))
     | _ -> go_children label depth children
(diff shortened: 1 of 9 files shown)


srk/src/syntax.ml
```diff
@@ -408,15 +408,15 @@ let substitute_sym srk subst sexpr =
           children
           Env.empty
       in
-      substitute srk (Env.find env) (subst k)
-      |> decapture srk 0 (depth - (List.length children))
+      let f (ind, typ) = try (Env.find env ind) with _ -> (mk_var srk ind typ) in
+      decapture srk (List.length children) (depth - (List.length children)) (subst k)
+      |> substitute srk f
```
zkincaid: Style: better to avoid giving non-descriptive names to short functions, and generic error capture:
```
      |> substitute srk (fun (ind, typ) ->
            try Env.find env ind with Not_found -> mk_var srk ind typ)
```

PR #24, opened 2021-04-19
arrays as first order types

JakeSilverman:
This PR adds arrays as a first order type to srk.

Two things are notable missing from this PR

1) We discussed including select atoms in coordinate_systems. This changes seems like it might greatly complicate current code given coordinate_systems reliance on QQVectors.

2) No current support for interpretation of arrays

### DIFF
--- a/srk/src/interpretation.ml
+++ b/srk/src/interpretation.ml
@@ -133,7 +133,7 @@ let rec evaluate_term interp ?(env=Env.empty) term =
       begin match Expr.refine interp.srk (unfold_app interp func args) with
         | `ArithTerm t ->
           evaluate_term interp ~env t
-        | `ArrTerm _ -> assert false
+        | `ArrTerm _
         | `Formula _ ->
           invalid_arg "evaluate_term: ill-typed function application"
       end
@@ -395,9 +395,8 @@ let select_implicant interp ?(env=Env.empty) phi =
 
 let destruct_atom srk phi =
   match Formula.destruct srk phi with
-  | `Atom (`Arith (op, s, t)) -> `Comparison (op, s, t)
-  | `Atom (`ArrEq _) ->  invalid_arg @@ Format.asprintf 
-      "destruct_atom: %a is syntactic sugar for quantified expression" (Formula.pp srk) phi
+  | `Atom (`Arith (op, s, t)) -> `Arith_Comparison (op, s, t)
+  | `Atom (`ArrEq (a, b)) ->  `Arr_Comparison (a, b)
   | `Proposition (`App (k, [])) ->
     `Literal (`Pos, `Const k)
   | `Proposition (`Var i) -> `Literal (`Pos, `Var i)
@@ -409,8 +408,8 @@ let destruct_atom srk phi =
     end
   | `Tru ->
     let zero = mk_real srk QQ.zero in
-    `Comparison (`Eq, zero, zero)
-  | `Fls -> `Comparison (`Eq, mk_real srk QQ.zero, mk_real srk QQ.one)
+    `Arith_Comparison (`Eq, zero, zero)
+  | `Fls -> `Arith_Comparison (`Eq, mk_real srk QQ.zero, mk_real srk QQ.one)
   | _ ->
     invalid_arg @@ Format.asprintf "destruct_atom: %a is not atomic" (Formula.pp srk) phi
 
--- a/srk/src/interpretation.ml
+++ b/srk/src/interpretation.ml
@@ -395,8 +395,8 @@ let select_implicant interp ?(env=Env.empty) phi =
 
 let destruct_atom srk phi =
   match Formula.destruct srk phi with
-  | `Atom (`Arith (op, s, t)) -> `Arith_Comparison (op, s, t)
-  | `Atom (`ArrEq (a, b)) ->  `Arr_Comparison (a, b)
+  | `Atom (`Arith (op, s, t)) -> `ArithComparison (op, s, t)
+  | `Atom (`ArrEq (a, b)) ->  `ArrEq (a, b)
   | `Proposition (`App (k, [])) ->
     `Literal (`Pos, `Const k)
   | `Proposition (`Var i) -> `Literal (`Pos, `Var i)
@@ -408,8 +408,8 @@ let destruct_atom srk phi =
     end
   | `Tru ->
     let zero = mk_real srk QQ.zero in
-    `Arith_Comparison (`Eq, zero, zero)
-  | `Fls -> `Arith_Comparison (`Eq, mk_real srk QQ.zero, mk_real srk QQ.one)
+    `ArithComparison (`Eq, zero, zero)
+  | `Fls -> `ArithComparison (`Eq, mk_real srk QQ.zero, mk_real srk QQ.one)
   | _ ->
     invalid_arg @@ Format.asprintf "destruct_atom: %a is not atomic" (Formula.pp srk) phi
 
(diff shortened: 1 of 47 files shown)


srk/src/quantifier.ml
```diff
@@ -238,20 +240,21 @@ let mbp_virtual_term srk interp x atoms =
    - psi is negation- and quantifier-free formula, and contains no free
      variables
    - every atom of in psi is either a propositial variable or an arithmetic
-     atom of the form t < 0, t <= 0, or t = 0
+     atom of the form t < 0, t <= 0, or t = 0 or array equality atom (meaning
```
zkincaid: Remove parenthetical

srk/src/srkSmtlib2.ml
```diff
@@ -137,7 +137,8 @@ module MakeSMT2Srk(C : sig type t val context : t Syntax.context end) = struct
   type 'a gexpr = ('a, Syntax.typ_fo) Syntax.expr
   let rec expr_of_smt of_symbol t =
     let formula_of = formula_of_smt of_symbol in
-    let term_of = term_of_smt of_symbol in
+    let aterm_of = aterm_of_smt of_symbol in
```
zkincaid: should be arith_term_of

srk/src/srkSmtlib2.ml
```diff
@@ -397,7 +404,7 @@ module MakeSMT2Srk(C : sig type t val context : t Syntax.context end) = struct
   (* functions are debruijn indexed *)
   let model_of_smt of_symbol (m : model) =
     let open Syntax in
-    let term_of_smt of_sym t = term_of_smt of_sym t in
+    (*let term_of_smt of_sym t = term_of_smt of_sym t in*)
```
zkincaid: delete

srk/src/interpretation.ml
```diff
@@ -131,8 +131,9 @@ let rec evaluate_term interp ?(env=Env.empty) term =
     | `App (k, []) -> real interp k
     | `App (func, args) ->
       begin match Expr.refine interp.srk (unfold_app interp func args) with
-        | `Term t ->
+        | `ArithTerm t ->
           evaluate_term interp ~env t
+        | `ArrTerm _ -> assert false
```
zkincaid: ArrTerm/Formula cases should be consistent.   Assert false ok

srk/src/interpretation.ml
```diff
@@ -149,6 +150,7 @@ let rec evaluate_term interp ?(env=Env.empty) term =
       raise Divide_by_zero
     | `Binop (`Div, dividend, divisor) -> QQ.div dividend divisor
     | `Binop (`Mod, t, modulus) -> QQ.modulo t modulus
+    | `Select _ -> assert false
```
zkincaid, line 153: Why can't we evaluate select?

srk/src/interpretation.ml
```diff
@@ -253,6 +256,7 @@ let select_implicant interp ?(env=Env.empty) phi =
         | `Neg -> mk_neg srk t_term
       in
       (term, t_impl)
+    | `Select _ -> assert false
```
zkincaid, line 259: Why can't we find an implicant involving select?

srk/src/interpretation.ml
```diff
@@ -389,7 +395,9 @@ let select_implicant interp ?(env=Env.empty) phi =
 
 let destruct_atom srk phi =
   match Formula.destruct srk phi with
-  | `Atom (op, s, t) -> `Comparison (op, s, t)
+  | `Atom (`Arith (op, s, t)) -> `Comparison (op, s, t)
+  | `Atom (`ArrEq _) ->  invalid_arg @@ Format.asprintf
```
zkincaid: I think this is unexpected -- can just add an array equality to the return type.

srk/src/syntax.mli
```diff
@@ -346,17 +428,28 @@ val mk_exists_consts : 'a context -> (symbol -> bool) -> 'a formula -> 'a formul
...
    symbols or quantifiers. *)
 val eliminate_ite : 'a context -> 'a formula -> 'a formula
 
+(** Given a formula [phi], compute an equivalent formula without
+   arr_eq terms. *)
+val eliminate_arr_eq : 'a context -> 'a formula -> 'a formula
+
+(** Given a formula [phi], compute an equivalent formula without
+   array stores. Note that this implicitly eliminates arr_eqs too.*)
+val eliminate_stores : 'a context -> 'a formula -> 'a formula
```
zkincaid: For parallelism, should just be "eliminate_store".  Why should this also eliminate array equalities?

srk/src/syntax.ml
```diff
@@ -1308,6 +1643,54 @@ let eliminate_ite srk phi =
...
+    | `Atom (`ArrEq (s, t)) ->
+      let s_i = mk_select srk (decapture srk 0 1 s) (mk_var srk 0 `TyInt) in 
+      let t_i = mk_select srk (decapture srk 0 1 t) (mk_var srk 0 `TyInt) in
+      mk_forall srk ~name:"i" `TyInt (mk_eq srk s_i t_i)
+    | phi -> Formula.construct srk phi
+  in
+  Formula.eval srk alg phi
+
+let eliminate_stores srk phi =
+  let mk_op op =
```
zkincaid: promote mk_op to top-level (maybe call it mk_compare)?

srk/src/syntax.mli
JakeSilverman: I am going to delete this function for the time being. Really it belongs in pmfa.ml. It eliminates stores involved in array equality and selects (but not those in say, an uninterpreted function or proposition). In the equality case it is doing so by expanding the array equality (hence why the array equality is eliminated).

srk/src/interpretation.ml
JakeSilverman: W.r.t. finding implicant involving/evaluating select: during our previous discussion I thought we were not happy with my definition of concrete arrays. You will notice that the type 'a value doesn't support arrays and thus we cannot currently handle interpretations with array variables. It follows that we can't evaluate select and thus we also can't find an implicant involving select.

I am happy to bring the previous definition back (with modifications). If I recall, there were two issues with the previous definition: (1) our language has the sort integer-arrays but I was interpreting them using arrays in which the elements are rational numbers. (2) I had a notion of extensional equality.

W.r.t. 1: I did this because we are also mapping integer sorted variables to rational numbers. I did this to follow that same convention but I am just as happy to use integers here. Let me know which you think is best.

W.r.t. 2: I will just remove any notion of equality for the concrete arrays.

srk/src/interpretation.ml
JakeSilverman: I'll keep the current error message (and just have them both point use that same error).

srk/src/interpretation.ml
JakeSilverman: I added this in. It is unclear to me if the changes I made to Quantifier.ml are correct. I more or less just copied the literal case to the "Arr_Comparison" case (except during testing if presburger), but I should really just be using "assert false" during the function related to mbp for the "Arr_Comparison" case.

zkincaid: ArrEq is a better name than Arr_Comparison.  It's a particular type of comparison, and this project uses camel case for type constructors rather than underscores.

srk/src/interpretation.ml
zkincaid: Actually, I'm ok with leaving it out.  The choice of not implementing functionality that you don't currently need is reasonable.
(The larger issue with the previous implementation is that it made an arbitrary choice that arrays are piecewise constant functions.  The implementation of function values in interpretations gives a more  general solution, in which function values are expressions with parameters -- so any [definable](https://en.wikipedia.org/wiki/Definable_set) function has a representation as a function value.)

PR #25, opened 2021-04-28
Support for sup-lin chc. Removed most normalization from srkZ3 parsin…

JakeSilverman:
…g. QLIALift. Various chc helper functions

Note that this push add a few "TODOS" for locations that need to be touched up after the arrays as first order types is pushed. Additionally, it adds a "construct" function to the Formula module that duplicates a function that already exists in the arrays as first order types PR.

### DIFF
--- a/srk/src/chc.ml
+++ b/srk/src/chc.ml
@@ -1,414 +1,381 @@
 (* Constrained horn clauses *)
 open Syntax
-open BatPervasives
+(*open BatPervasives*)
 module DynArray = BatDynArray
 module D = Graph.Pack.Digraph
 module WG = WeightedGraph
 
-type relation = int
-type rel_atom = relation * symbol list
-type 'a fp = { mutable rules : (rel_atom list * 'a formula * rel_atom) list; 
-               mutable queries : relation list;
-               rel_ctx : (string * typ_fo list) DynArray.t } 
+type proposition = { symbol : symbol; names : string list }
+module Proposition = struct
+  (* Relation atoms have as parameters free variables. We maintain the invariant
+   * that these free variables be sequential. The int option in this type
+   * describes the debruijn index of the first parameter; it is `None` in the
+   * case that the atom has no parameters.*)
+  type t = proposition
+ 
+  let symbol_of prop  = prop.symbol
+  let names_of prop = prop.names
 
 
-let mk_relation fp ?(name="R") typ =
-  DynArray.add fp.rel_ctx (name, typ);
-  DynArray.length fp.rel_ctx - 1
-let type_of fp rel = snd (DynArray.get fp.rel_ctx rel)
-let name_of fp rel = fst (DynArray.get fp.rel_ctx rel)
+  let typ_of_params srk prop =
+    match typ_symbol srk prop.symbol with 
+    | `TyBool 
+    | `TyFun ([], `TyBool) -> []
+    | `TyFun (lst, `TyBool) -> lst
+    | _ -> assert false
 
-let rel_of_atom (relation, _) = relation
-let params_of_atom (_, params) = params
+  let prop_of srk prop init_index =
+    let params = 
+      BatList.mapi (fun ind typ ->
+          mk_var srk (ind + init_index) typ)
+        (typ_of_params srk prop)
+    in
+    mk_app srk (symbol_of prop) params
+
+  let mk_proposition srk symbol names =
+    match typ_symbol srk symbol with 
+    | `TyBool when BatList.is_empty names -> { symbol; names}
+    | `TyFun (typs, `TyBool) when List.length names = List.length typs ->
+      { symbol; names}
+    | _ -> invalid_arg "mk_proposition: ill-formed proposition"
+end
+
+ type 'a fp = { rules : (proposition * proposition list * 'a formula) list; 
+                queries : Symbol.Set.t }
 
 let typ_symbol_fo srk sym =
   match typ_symbol srk sym with
   | `TyInt -> `TyInt
   | `TyReal -> `TyReal
   | `TyBool -> `TyBool
+  | `TyArr -> `TyArr
   (* TODO: arrays *)
   | _ -> assert false
-
-let mk_rel_atom srk fp rel syms =
-  if ((type_of fp rel) = (List.map (typ_symbol_fo srk) syms))
-  then rel, syms
-  else invalid_arg "Type Error mk_rel_atom"
-
-module Relation = struct
-  type t = relation 
-  module Set = SrkUtil.Int.Set 
-  let compare = Stdlib.compare
-
-  let pp fp formatter rel =
-    Format.fprintf formatter "@%s:R%n@" (name_of fp rel) rel
-  let show fp = SrkUtil.mk_show (pp fp)
-end
-
-let pp_rel_atom srk fp formatter (rel, symbols) =
-    Format.fprintf formatter "%a(@[%a@])"
-      (Relation.pp fp) rel
-      (SrkUtil.pp_print_enum_nobox (pp_symbol srk)) (BatList.enum symbols)
-
-let show_rel_atom srk fp = 
-  SrkUtil.mk_show (pp_rel_atom srk fp)
 
 module Fp = struct
   type 'a t = 'a fp
 
-  let pp_hypothesis srk fp formatter (rel_atoms, phi) =
-    Format.fprintf formatter "(@[";
-    SrkUtil.pp_print_enum
-      ~pp_sep:(fun formatter () -> Format.fprintf formatter "@ /\\ ")
-      (pp_rel_atom srk fp)
-      formatter
-      (BatList.enum rel_atoms);
-    Format.fprintf formatter "@]/\\%a)" (Formula.pp srk) phi
-  let pp_rule srk fp formatter (rel_atoms, phi, conc) =
-    Format.fprintf formatter "(@[%a@ => %a@])"
-      (pp_hypothesis srk fp) (rel_atoms, phi)
-      (pp_rel_atom srk fp) conc
+  let pp_rule srk formatter (conc, hypo_props, phi) =
+    let body, _ =
+      List.fold_left (fun (phi, counter) prop -> 
+          mk_and srk [Proposition.prop_of srk prop counter; phi], 
+          counter + List.length (Proposition.typ_of_params srk prop)) 
+        (phi, List.length (Proposition.typ_of_params srk conc))
+        hypo_props
+    in
+    let pre_quantified = mk_if srk body (Proposition.prop_of srk conc 0) in
+    let phi = 
+      List.fold_left (fun phi prop ->
+          BatList.fold_left2 (fun phi typ name ->
+              mk_forall srk ~name:name typ phi)
+            phi
+            (Proposition.typ_of_params srk prop)
+            (Proposition.names_of prop))
+        pre_quantified
+        (conc :: hypo_props)
+    in
+    Formula.pp srk formatter phi
 
   let pp srk formatter fp = 
     Format.fprintf formatter "(Rules:\n@[";
     SrkUtil.pp_print_enum
       ~pp_sep:(fun formatter () -> Format.fprintf formatter "@ \n ")
-      (pp_rule srk fp)
+      (pp_rule srk)
       formatter
       (BatList.enum fp.rules);
     Format.fprintf formatter "@]\nQueries:@[%a@])"
-      (SrkUtil.pp_print_enum_nobox (Relation.pp fp)) 
-      (BatList.enum fp.queries)
+      (SrkUtil.pp_print_enum_nobox (pp_symbol srk)) 
+      (Symbol.Set.enum fp.queries)
 
   let show srk = SrkUtil.mk_show (pp srk)
 
-  let create () = {rules=[];queries=[];rel_ctx=(DynArray.create ())}
+  let empty = {rules=[];queries=Symbol.Set.empty}
 
-  let add_rule fp hypo phi conc = fp.rules <- (hypo, phi, conc) :: fp.rules 
+  let add_rule fp conc hypo phi = { fp with rules = (conc, hypo, phi) :: fp.rules} 
 
-  let add_query fp query = fp.queries <- query :: fp.queries 
-
-  let get_active_relations fp =
-    List.fold_left
-      (fun rset (hypo_atoms, _, conc_atom) ->
-         List.fold_left
-           (fun rset rel_atom -> Relation.Set.add (rel_of_atom rel_atom) rset)
-           (Relation.Set.add (rel_of_atom conc_atom) rset)
-           hypo_atoms)
-      (List.fold_left 
-         (fun rset query -> Relation.Set.add query rset)
-         Relation.Set.empty
-         fp.queries)
-      fp.rules
-
-  let get_relations_declared fp =
-    BatList.of_enum (0 -- ((DynArray.length fp.rel_ctx) - 1))
-
-  let is_linear rules =
-    BatList.fold_left 
-      (fun acc (hypo_atoms, _, _) -> acc && (List.length hypo_atoms) <= 1)
-      true
-      rules
-
-  let mk_eq_by_sort srk s1 s2 =
-    assert (typ_symbol srk s1 = typ_symbol srk s2);
-    match typ_symbol_fo srk s1 with
-    | `TyInt | `TyReal -> mk_eq srk (mk_const srk s1) (mk_const srk s2)
-    | `TyBool -> mk_iff srk (mk_const srk s1) (mk_const srk s2)
-
-  let normalize srk fp =
-    let rules' =
-      List.map (fun (hypos, constr, conc) ->
-          let update_atom (atom, eqs, syms) =
-            let params, eqs, syms =
-              List.fold_right
-                (fun param (params, eqs, syms) ->
-                   if not (Symbol.Set.mem param syms) then
-                     param :: params, eqs, (Symbol.Set.add param syms)
-                   else
-                     let s = dup_symbol srk param in
-                     let eq = mk_eq_by_sort srk s param in
-                     s :: params, eq :: eqs, Symbol.Set.add s syms)
-                (params_of_atom atom) 
-                ([], eqs, syms)
-            in
-            let atom' = mk_rel_atom srk fp (rel_of_atom atom) params in
-            atom', eqs, syms
-          in
-          let (hypos', eqs, syms) =
-            List.fold_left 
-              (fun (hypos', eqs, syms) hypo ->
-                 let atom', eqs', syms' = update_atom (hypo, eqs, syms) in
-                 atom' :: hypos', eqs', syms')
-              ([], [],  Symbol.Set.empty)
-              hypos
-          in
-          let (conc', eqs, syms) = update_atom (conc, eqs, syms) in
-          let constr' = 
-            mk_exists_consts 
-              srk 
-              (fun s -> Symbol.Set.mem s syms) 
-              (mk_and srk (constr :: eqs))
-          in
-          hypos', constr', conc')
-        fp.rules
-    in
-    {rules=rules'; queries=fp.queries; rel_ctx=fp.rel_ctx} 
-
+  let add_query fp query = { fp with queries = Symbol.Set.add query fp.queries} 
 
   let goal_vert = -2 
   let start_vert = -1
-  let emptyarr = BatArray.init 0 (fun _ -> failwith "empty")
 
-  let linchc_to_weighted_graph srk fp pd =
-    let fp = normalize srk fp in
+  let linchc_to_weighted_graph srk (fp : 'a t) pd =
     let open WeightedGraph in
-    (* The edges of the graph are of the form [(syms', syms', constr)] where 
-     * [syms] are the var symbols used in the hypothesis relation, [syms'] are
-     * those used in the conclusion relation, and [constr] in the constraint
-     * whose free variables are limited to [syms union syms']. Further, [syms]
-     * and [syms'] are always disjoint. *)
+    (* The edges of the graph are of the form 
+     * [(conc params, hypo params, constr)]*) 
     let alg = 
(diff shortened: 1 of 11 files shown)


srk/src/transitionFormula.mli
```diff
@@ -42,6 +42,15 @@ val symbols : 'a t -> (symbol * symbol) list
 (** Retrieve the predicate that identifies skolem constants *)
 val exists : 'a t -> (symbol -> bool)
 
+(** Retrieve the underlying formula *)
+val update_formula : 'a t -> 'a formula -> 'a t
```
zkincaid, line 46: Documentation is wrong.

srk/src/iteration.ml
```diff
@@ -426,6 +426,43 @@ module NonlinearRecurrenceInequation = struct
...
+                else mk_const srk s)
+             phi')
+        in
+        if qt = `Forall then mk_not srk phi' else phi'
+      | open_form -> Formula.construct srk open_form
+    in
+    Formula.eval srk alg phi
+
+  let abstract srk tf =
+    let tf' = TF.update_formula tf (integer_qe_mbp srk (TF.formula tf)) in
```
zkincaid, line 459: Use TF.map_formula

srk/src/iteration.ml
```diff
@@ -426,6 +426,43 @@ module NonlinearRecurrenceInequation = struct
     |> mk_and srk
 end
 
+module QLIALift (A : PreDomain) = struct 
+  type 'a t = 'a A.t
+
+  let integer_qe_mbp srk phi =
```
zkincaid, line 432: This belongs in quantifier.ml.

srk/src/chc.mli
```diff
@@ -36,19 +36,41 @@ module Fp : sig
...
    * be reached in the fp where recursion over-approximated using the 
    * star operator of the provided predomain [pd] and returns no otherwise.*)
   val check : 
     'a context -> 'a fp -> (module Iteration.PreDomain) -> 
     [> `No | `Unknown | `Yes]
+  (** [query_vc_condition srk fp pd] returns the final vc condition used in
+   * [check srk fp pd]. That is, the vc condition to determine whether a query
+   * relation can be reached in the fp where recursion over-approximated using 
+   * the star operator of the provided predomain [pd].*)
+  val query_vc_condition :
```
zkincaid, line 36: Why is this exposed?

srk/src/chc.mli
```diff
@@ -5,14 +5,14 @@ type relation
 type rel_atom
 type 'a fp = {mutable rules : (rel_atom list * 'a formula * rel_atom) list;
```
zkincaid: Why is this type def exposed?

srk/src/chc.mli
```diff
@@ -36,19 +36,41 @@ module Fp : sig
   val add_rule : 'a fp -> rel_atom list -> 'a formula -> rel_atom -> unit
   (** Adds query to fp and returns a fresh name for query *)
   val add_query : 'a fp -> relation -> unit
-  val get_relations_used : 'a fp -> Relation.Set.t
+  (* Returns set of relations that occur in either a rule or query in fp *)
+  val get_active_relations : 'a fp -> Relation.Set.t
```
zkincaid: Why is this useful?

srk/src/chc.mli
```diff
@@ -36,19 +36,41 @@ module Fp : sig
...
+  (* Limit each variable symbol to occur at most once as a parameter
+   * in any given rule.
+   *
+   * For example, the rule R1(x) -> R2(x) is turned into R1(x) /\ x = x' => R2(x')
+   *
+   * Additionally quantifies over free vars in a constraint to that do not
+   * appear as a parameter to a relation atom in the rule that constraint 
+   * belongs to (that is, skolem constants will be unskomelized).
+   *)
+  val normalize : 'a context -> 'a fp -> 'a fp
```
zkincaid: Why is this exposed?

srk/src/chc.mli
```diff
@@ -36,19 +36,41 @@ module Fp : sig
...
+  (* Creates an equivalent fp in which boolean parameters have been mapped to
```
zkincaid: Not equivalent.  Equisatisfiable?

srk/src/chc.ml
```diff
@@ -91,27 +101,73 @@ module Fp = struct
       fp.rules
 
   let get_relations_declared fp =
-    BatList.of_enum (0 -- ((DynArray.length fp.rel_ctx) - 1)) 
+    BatList.of_enum (0 -- ((DynArray.length fp.rel_ctx) - 1))
 
-  let is_linear fp =
+  let is_linear rules =
```
zkincaid: BatList.for_all

srk/src/chc.ml
```diff
@@ -91,27 +101,73 @@ module Fp = struct
...
-      fp.rules
+      rules
+
+  let mk_eq_by_sort srk s1 s2 =
+    assert (typ_symbol srk s1 = typ_symbol srk s2);
+    match typ_symbol_fo srk s1 with
+    | `TyInt | `TyReal -> mk_eq srk (mk_const srk s1) (mk_const srk s2)
+    | `TyBool -> mk_iff srk (mk_const srk s1) (mk_const srk s2)
+
+  let normalize srk fp =
```
zkincaid: If you want the invariant that each rule has a particular form, might be better to maintain the invariant in add_rule instead of allowing the invariant to be broken and later restored by calling normalize.

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
         if is_zero x then y else if is_zero y then x
```
zkincaid, line 111: Move algebra operations to top level

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
...
+    BatArray.iter (fun v -> D.add_vertex graph v) verts;
+    BatList.iter (fun (hypo_atoms, _, conc_atom) ->
+        List.iter (fun hatom ->
+            D.add_edge 
+              graph 
+              (verts.(rel_of_atom hatom)) 
+              (verts.(rel_of_atom conc_atom)))
+          hypo_atoms)
+      fp.rules;
+    (* We create a new graph in which each scc is collapsed to
```
zkincaid, line 272: SCCs are [already](http://ocamlgraph.lri.fr/doc/Components.Make.html) ordered topologically, so no need to sort again.  Can check superlin property using scc number.

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
...
-    if is_linear fp then (
-      let wg = to_weighted_graph srk fp pd in
-      let _, _, phi = WeightedGraph.path_weight wg start_vert goal_vert in
-      begin match Smt.is_sat srk phi with
-        | `Unsat -> `No
-        | `Unknown -> `Unknown
-        | `Sat -> `Unknown
-      end)
-    else failwith "No methods for solving non lin fp"
+  let is_super_linear fp =
```
zkincaid: Doesn't return a boolean -- rename to "stratify" or something?

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
...
+                    let soln = substitute_const srk subst soln in
+                    hypo_atoms, mk_and srk [soln; constr])
+                  else (atom :: hypo_atoms, constr))
+                ([], constr)
+                hypo_atoms
+            in
+            hypo_atoms', constr', conc)
+            ruleset
+        in
+        if is_linear ruleset' then (
```
zkincaid: Don't need to check if it's linear

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
...
+                    let syms, soln = Hashtbl.find solution (rel_of_atom atom) in
+                    let subst =
+                      (fun sym ->
+                         mk_const
+                           srk
+                           (List.nth
+                              (params_of_atom atom)
+                              (BatArray.findi ((=) sym) syms)))
+                    in
+                    let soln = substitute_const srk subst soln in
```
zkincaid: Is the solution guaranteed to only use the symbols in syms?

srk/src/chc.ml
```diff
@@ -120,120 +176,312 @@ module Fp = struct
...
+                  else (atom :: hypo_atoms, constr))
+                ([], constr)
+                hypo_atoms
+            in
+            hypo_atoms', constr', conc)
+            ruleset
+        in
+        if is_linear ruleset' then (
+          let sub_fp = {rules=ruleset'; queries=[]; rel_ctx=fp.rel_ctx} in
+          let wg = linchc_to_weighted_graph srk sub_fp pd in
```
zkincaid: Just construct the weighted graph directly?

srk/src/chc.ml
```diff
@@ -244,120 +492,119 @@ module ChcSrkZ3 = struct
...
     | BOOL_SORT -> `TyBool
-    | _ -> failwith "TODO: allow function types"
+    (*| ARRAY_SORT -> `TyArr*)
+    |_ -> failwith "function types not currently supported in rel param"
 
-  (* Creates a relation atom from a z3 predicate in which each argument
-   * to the predicate is an integer. Replaces integer [i] with value
-   * located at key [i] in table [ind_to_sym] when such a key exists *)
-  let rel_atom_of_z3 srk fp ind_to_sym rsym_to_int names z3pred =
+  let rel_atom_of_z3 srk fp ind_to_sym rsym_to_int z3pred =
```
zkincaid: Why can't you use SrkZ3.formula_of_z3?

srk/src/chc.ml
```diff
@@ -244,120 +492,119 @@ module ChcSrkZ3 = struct
...
-        else BatHashtbl.add ind_to_sym ind sym)
-      fresh_index_map;
-    atom, !eq_syms
-
+    (relation, args), !eqs
 
   let parse_z3fp ?(z3queries=[]) srk fp z3fp =
     let rsym_to_int = BatHashtbl.create 91 in
     let decl_kind e = Z3.FuncDecl.get_decl_kind (Z3.Expr.get_func_decl e) in
     let parse_rule rule =
```
zkincaid: Can you use SrkZ3.formula_of_z3 and then manipulate the srk formula?

srk/src/chc.mli
JakeSilverman: I am worried about Z3s ability to handle the pmfa fragment and I don't want to assume that every chc is a pmfa chc. Basically, I want to do some post-processing in the pmfa-chc case that I don't want to do in the general chc case. Specifically, I want to convert the pmfa vc condition to an equisatisfiable lia vc condition and then run Z3 on the lia vc condition.

srk/src/chc.ml
JakeSilverman: This is reasonable. My intention here is that the client can have the chc be in whatever form they want, but there are some functions (like check/solve) that if they intend to call them, they ought to normalize first.

This normalization procedure does two things (1) it ensures that each constraint has no free vars other than those that occur as a relation argument (2) it ensures that each symbol is used at most once-per-rule as an argument to a relation. I do see value in chc without these constraints.

W.r.t. (1): it is much easier for me to compute the partitioning of the chc array variables for the pmfa offset partitioning if I do so without this restriction.

W.r.t. (2): I figure that allowing this is freedom when typing up chc is just to have. It also has the added bonus that the chc generated by parse_z3fp look ever so slightly more like the original smt2 format chc.

______

If we want to remove/hide normalization likely I will 
(1) add normalization invariant (2) listed above as an invariant to all fp via add_rule, as you suggest.
(2) Not require normalization invariant (1) as an invariant to all fp, but instead whenever check/solve is entered, create a new temporary fp that has no free vars other than those that occur as a relation argument and run the check/solve on this temporary fp.

srk/src/chc.ml
JakeSilverman: If my code is correct, that should be the case. "syms, soln" is the solution to a relation symbol where syms[n] is the symbol used for nth argument to that relation. Solve_linear_chc converts the chc to a graph where the nodes are relations and the edges are constraints whose free variables are limited to the symbols used as arguments to the input and output relation. This invariant is initially ensured by calling the normalization procedure and then is maintained through the add, mul, and star function of the algebra.

srk/src/chc.ml
JakeSilverman: I believe that should work. I would still need some kind of rel_atom_of_formula_predicate function though. This is because my relation atoms are much more restricted than those allowed in smt2, even than those produced by seahorn. I require that arguments to a relation symbol be variable symbols.

srk/src/chc.ml
JakeSilverman: As per previous note - yes. I believe that should be fine.

srk/src/chc.mli
JakeSilverman: See conversation below. The intention was that client can create a chc that isn't normal and then just normalize it when they desire. I am likely going to do what you said which is instead just some invariant to the fp object by modifying add_rule.

srk/src/chc.mli
JakeSilverman: I thought I was using it but clearly I am not. It is currently in use in my offset analysis, but can fairly easily be replaced. I will get rid of it.

srk/src/chc.mli
zkincaid: Can't you just call "solve" on the goal relation to fetch the formula?

srk/src/chc.ml
zkincaid: How is it easier to compute the partition *without* the restriction?  The inputs to your partitioning algorithm are without normalization are a superset of the un-normalized inputs, no?

More broadly -- it seems there are (potentially) two different purposes here: is this module intended to be a representation of the syntax of a system of CHCs, or is it supposed to be a solver interface?  If it's the latter, the only thing that should be exposed are operations for constructing and solving CHCs.  If the solver requires normalization, everything should be normalized (transparently from the user's perspective).  If it's the former, a different set of operations ought to be exposed (and "normalize" may indeed be one of them).  If you want both, perhaps there should be two modules instead of one.

JakeSilverman: To answer the second question:
The answer is both. I agree splitting probably makes sense here. Perhaps a module "Fp" that lives in "Chc.ml" which has submodules "Syntax" "Solve" and "SrkZ3". Potentially going a bit crazy with the nesting of modules here though.

The "normalization" function will then live in the "Solve" module (I can also just hide the procedure from the user and just run the procedure whenever solve/check is called).

______

To answer your first question: I was under the impression that you are suggesting the CHC module ensures that every CHC that is created fulfills these two requirements. Yes the offset analysis deals with normalized and un-normalized inputs, but the offset analysis proceeds by first normalizing chc in a way different than the normal form above (and most importantly not compatible with the one above). The offset-normal form has the same invariant (2) but invariant (1) is that there are no existential quants; every existential quant is skolemized.  Requiring that all chc have this normal form would mean that I cannot modify chc to be of the offset-normal form in the offset analysis and it makes the offset analysis more complicated to implement.

srk/src/chc.mli
JakeSilverman: Yeah I could do that. I will need to expose the goal relation though, which is currently hidden from the client. The current approach disallows the client from doing anything weird with the goal relation (i.e., adding a rule that has the goal relation in the atom).

srk/src/chc.ml
zkincaid: You don't need an Fp module to contain everything --- everything is already contained in the file-level module Chc.

The normalization function should live in syntax, because it manipulates syntax.  The Solver module should contain logic for solving systems of CHCs -- it should call normalize in the syntax module if it needs to.   Since you want to have a module for representing syntax, the syntax should match expectations for CHCs.  The only constant symbols I would expect to see are the relation symbols -- everything else out to be a variable, implicitly universally quantified.  The decision to use constants everywhere (and to have a new type for relations rather than re-using symbols) was based on arrays not being first-order types.  If these details are hidden from clients of the module then it's not much of an issue, but if the syntax is accessible from outside this module then more care needs to be taken to make the syntax make sense.

JakeSilverman: Hmm. I forgot why relations were a different type... maybe it makes since to switch them back to symbols now that arrays as FO types. Though this version is perhaps more future proof since it can easily be changed to handle function types (note that support for this is currently off).

_______

The current syntax of a rule is:
hypothesis: a list of relation atoms
constraint: a formula with no fv.
conclusion: a relation atom.

Let syms_constr be the list of constant symbols that occur in constraint but are not an argument to any relation atom. Let syms_atoms be the list of constant symbols that occur as an argument to a relation atom.

We interpret a rule as the following formula: \forall syms_atoms . ((\bigand_{hypo in hypothesis} hypo) /\ exists syms_constr. constraint)) => conclusion)

Perhaps this syntax isn't super intuitive - I can see why one would expect I use free_vars rather than symbols. My decision to use constant symbols rather than free_vars was that I just find constant symbols in general to be easier to work with.



______

If I make the switch from relations as their own types to symbols with func types, there is also a question of if I really even ought to have a special type for rule. Z3 encodes a rule as a formula. We definitely could just encode a rule as a formula to and throw some error if the formula doesn't look like a chc should. On one hand, I'm inclined to move closer to Z3s encoding because probably they put a lot of time thinking into their design decisions. But on the other hand I like not having the parse the rule each time to pull out the relation atoms in the formula.

zkincaid: Z3's design sacrifices type safety for convenience -- (nearly) everything is of type expression.  It's a reasonable compromise.  Providing type safety and convenient accessors by having a specialized type for horn clauses is also reasonable.

One could have a system of constraints like:
```
(declare-const k Int)
(declare-fun I (Int) Bool)
(assert (I 0))
(assert (forall ((x Int) (y Int)) (=> (and (I x) (< x 10) (= y (+ x k))) (I y))))
(assert (not (I 11)))
(check-sat)
(get-model)
```
Z3 happily computes a model for this:
```
(model 
  (define-fun k () Int
    (- 4))
  (define-fun I ((x!0 Int)) Bool
    (or (not (<= (- 2) x!0)) (and (<= (- 2) x!0) (<= 0 x!0) (not (<= 2 x!0)))))
)
```
My understanding of your representation of this system changes its semantics -- k is a constant symbol that doesn't appear in a relation atom, and so it's treated as an existentially quantified variable (in a negative position, so effectively universally quantified). Using variables would avoid this, and clarify the intended interpretation of the clause to the client.  This issue also comes up in transition formulas -- this is solved (inelegantly) by attaching an "exists" predicate to the transition formula that indicates which constant symbols are to be interpreted as constants and which as skolem constants.  This has been a source of confusion, so ideally I'd like to avoid it.

Is there something that's particularly difficult about manipulating variables?  Can we make it easier?

JakeSilverman: I see. Your understanding is correct - the scope of a symbol is in effect limited to a single rule and thus you can't have some symbol that denotes the same constant across multiple rules. It seems to me that what you're proposing is equivalent to adding support for existential horn clauses, which is perhaps something we want to do, but also something I recall us for whatever reason actively not pursuing during the initial design of CHC.ml.

My annoyance with manipulating variables is that I am in love with Hashtbls (feels wrong to admit I love imperative constructs) but it doesn't really make great sense to use Hashtbls for debruijn indices because a single variable may have a different index depending on where it occurs in the formula. To give a concrete example, look at the "offset_partitioning" function in https://github.com/JakeSilverman/duet/blob/pmfa/srk/src/pmfa.ml. I have some Hashtbl from variable symbol to equivalence class, and when I want to merge variable the equiv classes of symbol a and b I just look up their cell in the Hashtbl and merge the result. The debruijn index/more functional friendly approach would be to maintain a separate map for each expression and then merge the maps (probably still using my imperative BatUrefs), but I love Hashtbls and I don't want to give them up.

zkincaid: I don't mind if we don't support constants, but we shouldn't produce an incorrect result if we happen to run on a system that uses them.  This may introduce subtle bugs. For example, some iteration operators make use of symbols that are not to be treated as existentials (for example, the symbols might be represent exponential or logarithm functions, operational calculus operators, etc).  If the superlinear solver solves a linear system at some stratum and substitutes the solution for a relation into the constraint at the next stratum, are those symbols now existentially quantified?

There's nothing wrong with hash tables.  Offset partitioning is going to be imperative anyway due to the union-find.  To my eye, it would be simpler to avoid them in your offset_partitioning example: first you skolemize so that each variable reference is replaced by a unique symbol, then you use a global hash table to associate the unique symbols with union-find cells.  You could just cut out the middle man and associate variables with union-find cells, in essentially the same way that you're implementing "skolemize".  (By the way: you can implement maps over de Bruijn indicies is with [environments](https://github.com/zkincaid/duet/blob/modern/srk/src/syntax.mli#L14), which offer constant-time push and log-time find).  That is, you traverse the formula from the top down, and push a fresh union-find cell to the environment for each quantifier.  Every time you see a select, say
 (select (var 3) (var 0)), find the cell for var 3 and var 0 in the environment and merge them.  (Conceptually, var 0 shouldn't really be associated with a cell since it's of sort int.  But this is an easy trick to implement "all array symbols below a universal are merged together").

JakeSilverman: Yes. If we have the CHC

Rule 1: x = y => R1()

Rule 2: R1() /\ x !=y => R2()

This is interpreted as 

Rule 1: (\exists x, y. x = y) => R1()

Rule2: \exists x, y. (R1() /\ x !=y) => R2()

And thus the solution to R2 is  \exists x, y, x', y'.  x' = y' /\ x !=y

This was by design (that is, I don't consider this to be a bug). But that is because in the initial design we didn't support constant symbols. 

I am happy to change this design. I'm kind of wavering between the two different design choices
Choice 1:
(1) the relation atoms have as parameters free variables
(2) every variable symbol that occurs in a rule is a constant
(3) every free variable is implicitly universally quantified in the positive position

versus choice 2:
(1) the relation atoms have as parameters variable symbols
(2) every variable symbol that occurs in a rule that is not an argument to one of the relation atoms in this rule is a constant symbol
(3) the constraint may have additional free variables and these are implicitly existentially quantified (in the negative position)

Probably going to go with the first one since it seems cleaner (but the second one will require less changes to current code).

PR #26, opened 2021-08-16
Dta primal

happypig95:
Not sure if we want to merge everything here into master since some are specifically created for CAV artifact requirements. Let me know if you think some file should not appear in the master.


bench/tasks/bitprecise.set
```diff
@@ -32,7 +32,6 @@ bitprecise/Singapore-2.yml
 bitprecise/Stockholm-1.yml
 bitprecise/SyntaxSupportPointer01-1.yml
 bitprecise/SyntaxSupportPointer01-2.yml
-bitprecise/SyntaxSupportPointer01-3.yml
```
zkincaid, line 35: Was this benchmark proved unsafe by uautomizer?

bench/tools/cpachecker.py
```diff
@@ -0,0 +1,231 @@
+# This file is part of BenchExec, a framework for reliable benchmarking:
```
zkincaid: Is this file necessary?  BenchExec distribution should already have this file.

srk/src/terminationDTA.ml
```diff
@@ -440,14 +412,58 @@ module XSeq = struct
...
+        (mk_mod srk t (mk_real srk (QQ.of_int divisor)))
+        (mk_zero srk)
     in
     Periodic.map mk_divides dividend_xseqs
 
   let seq_of_notdivides_atom srk zz_divisor dividend_vec exp_poly abstraction =
     seq_of_divides_atom srk zz_divisor dividend_vec exp_poly abstraction
     |> seq_not srk
   
+  let omega_dom_basis_mat dim omega_dom_mat =
```
zkincaid, line 343: Need some comments describing the specification of the functions in this block

zkincaid (commented):
This pull request has conflicts in srk/src/terminationDTA.ml -- please pull in the latest changes first.

bench/tasks/bitprecise.set
happypig95: Yes this is the problematic one we decided to exclude by the submission time.

bench/tools/cpachecker.py
happypig95: I think the BencExec's version should be good.

PR #33, opened 2022-05-05
refactoring and bug fixes

happypig95:
refactoring to use polynomialUtil rather than pvutil, project out introduced symbols when eliminating floor div mod


srk/src/polynomialCone.ml
```diff
@@ -126,47 +118,20 @@ let add_polys_to_cone pc zero_polys nonneg_polys =
...
   Log.errorf "computing projection of cone part";
   let cone_generators = QQXs.one :: cone_generators in
-  let ctx = PVCTX.mk_context ~fix_const_dim:false Monomial.degrevlex cone_generators in
+  let ctx = PVCTX.mk_context Monomial.degrevlex cone_generators in
   let dim = PVCTX.num_dimensions ctx in
   Log.errorf "num dimension is: %d" dim;
   let polyhedron_rep_of_cone = polyhedron_of_cone cone_generators ctx in
   Log.errorf "polyhedron of cone before proj: %a" (Polyhedron.pp pp_dim) polyhedron_rep_of_cone;
-  let elim_dims = (0 -- (dim - 1)) |> BatEnum.filter
+  let elim_dims = (1 -- (dim - 1)) |> BatEnum.filter
```
zkincaid, line 134: Why starting at 1 instead of 0?

happypig95: Now monomial 1 is mapped to dimension -1 in polynomialUtil. And other monomials start from dimension 1 (rather than 0). So I think dimension 0 is not mapped to any monomial currently (indeed it will raise the exception if we start from 0)

PR #34, opened 2022-05-26
Fix README.md; OCaml >= 4.10 is necessary

ushitora-anqou:
To build duet, I needed to use OCaml 4.10 or newer because it uses `List.concat_map`, which has been supported since OCaml 4.10.


zkincaid:
Thanks!

PR #38, opened 2022-07-02
Cone intersection with affine subspace

nclskoh:
Fixed bug concerning taking the intersection by extracting affine polynomials; should be done using polyhedral projection instead.


srk/src/polynomialConeCpClosure.ml
```diff
@@ -188,79 +196,132 @@ let compute_transformation affine_basis ctxt : transformation_data =
...
+    (Polyhedron.pp pp_dim) p;
+  p
+
+(**
+    Given a regular polynomial cone C and a subspace U spanned by
+    [variables] and 1, such that the monomial ordering of C is a
+    graded elimination order that retains [variables] (e.g., devreglex),
+    compute the polyhedron [C \cap U].
+ *)
+let intersect_cone_and_affine_space cone variables =
```
zkincaid: This function is "PolynomialCone.restrict (fun m -> Monomial.degree m <= 1) cone", except that for non-proper cones you get the zeros [1; x; y; z] instead of just [1].  Do you need that behavior?

PR #50, opened 2022-10-18
PolynomialLattice

nclskoh:
Cleaned up polynomial lattice and added intersection.

### DIFF
--- a/srk/src/polynomialLattice.ml
+++ b/srk/src/polynomialLattice.ml
@@ -45,9 +45,7 @@ let pp pp_dim fmt t =
        ideal: @[<v 0>%a@]
      }"
     (Format.pp_print_list ~pp_sep:(fun fmt () -> Format.fprintf fmt "; ")
-       (QQXs.pp pp_dim))
-    (List.map (LinearQQXs.sparsify t.affine_context)
-       (IntLattice.basis t.affine_lattice))
+       (QQXs.pp pp_dim)) (affine_basis t)
     (Ideal.pp pp_dim) (ideal t)
 
 let reduce red polys =
@@ -58,6 +56,13 @@ let reduce red polys =
     polys
 
 let make ideal affine_polys : t =
+  let pp_dim = (fun fmt dim -> Format.fprintf fmt "%d" dim) in 
+  L.logf ~level:`always
+    "PolynomialLattice: of ideal @[%a@] and affine polynomials @[%a@]@;"
+    (Ideal.pp pp_dim) ideal
+    (Format.pp_print_list ~pp_sep:(fun fmt () -> Format.fprintf fmt "; ")
+       (QQXs.pp pp_dim))
+    affine_polys;
   let ideal = ideal in
   let affine_polys = reduce (Ideal.reduce ideal) affine_polys in
   let affine_context = LinearQQXs.min_context (BatList.enum affine_polys) in
--- a/srk/src/polynomialLattice.ml
+++ b/srk/src/polynomialLattice.ml
@@ -34,7 +34,7 @@ let make_context monomial_order polys =
   (LinearQQXs.make_context sorted_monos, sorted_monos)
 
 let ideal t = t.ideal
-
+  
 let affine_basis t =
   List.map (LinearQQXs.sparsify_affine t.affine_context)
     (IntLattice.basis t.affine_lattice)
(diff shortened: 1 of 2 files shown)


srk/src/polynomialLattice.ml
```diff
@@ -1,53 +1,60 @@
 open Polynomial
+open BatPervasives
 
 module L = Log.Make(struct let name = "srk.polyLattice" end)
 
+type context =
+  { ctx : LinearQQXs.context
+  ; monomials : Monomial.t list
```
zkincaid: Why do we need monomials?

srk/src/polynomialLattice.ml
```diff
@@ -1,53 +1,60 @@
...
 module MonomialSet =
   BatSet.Make(struct type t = Monomial.t
-                     let compare x y = match Monomial.degrevlex x y with
+                     let compare x y = match monomial_order x y with
                        | `Lt -> -1
                        | `Eq -> 0
                        | `Gt -> 1
               end)
 
-(* TODO: Check if we can disregard monomial order and just use [min_context].
```
zkincaid, line 25: The order doesn't matter, does it?

srk/src/polynomialLattice.ml
```diff
@@ -70,15 +78,16 @@ let make ideal affine_polys : t =
   let ideal = Rewrite.mk_rewrite monomial_order (Ideal.generators ideal) in
```
zkincaid: This is only a Groebner basis of "ideal" is already a Groebner basis wrt monomial_order.  Does monomial ordering actually matter?

srk/src/polynomialLattice.ml
```diff
@@ -115,28 +124,29 @@ let intersect t1 t2 =
...
     let weighted_affine1, weighted_affine2 = (weight w affine1, weight w' affine2) in
-    let weighted_ideal =
-      Rewrite.mk_rewrite elim_order (List.append weighted_ideal1 weighted_ideal2) in
-    let weighted_affine = reduce weighted_ideal
-                            (List.append weighted_affine1 weighted_affine2) in
-    let affine_context =
-      PolyVectorContext.mk_context ~increasing:true elim_order weighted_affine in
+    let weighted_ideal = Rewrite.mk_rewrite elim_order (weighted_ideal1 @ weighted_ideal2) in
+    let weighted_affine = reduce weighted_ideal (weighted_affine1 @ weighted_affine2) in
+    let affine_context = make_context weighted_affine in
```
zkincaid: make_context sorts monomials in degrevlex order, but they need to be sorted in elim_order for this to work.

srk/src/polynomialLattice.ml
zkincaid: Can expose an enumeration function in DenseConversion if you like

srk/src/polynomialLattice.ml
```diff
@@ -1,53 +1,60 @@
...
+              |> MonomialSet.elements
   in
-  LinearQQXs.make_context (MonomialSet.elements monos)
-
-(*
+  { ctx = LinearQQXs.make_context monos
+  ; monomials = monos
+  }  
+  
 let zero =
```
zkincaid: Is this empty or {0}?

srk/src/polynomialLattice.ml
```diff
@@ -153,12 +163,11 @@ let intersect t1 t2 =
...
     ; affine_context
     }
- *)
 
 let subset t1 t2 =
   let aff1 = affine_basis t1 in
   let ideal1 = ideal t1 in
-  let polys1 = List.append (Rewrite.generators ideal1) aff1 in
+  let polys1 = (Rewrite.generators ideal1) @ aff1 in
   List.for_all (fun p -> member p t2) polys1
```
zkincaid: Not sufficient --  e.g., imagine taking t2 to be the lattice w/ affine basis polys1.  Line 171 succeeds, but t2 doesn't (necessarily) contain all of ideal1, just its generators.

srk/src/polynomialLattice.ml
nclskoh: The reason for monomials and taking care of monomial ordering above was to make degrevlex here work, with the idea that the fresh dimension is larger than the rest. But I forgot about the graded part, and having all that extra stuff makes it unnecessarily abstruse. Dropped the above, and using elimination order to sort the monomials here.

srk/src/polynomialLattice.ml
nclskoh: {0}; I believe Ideal.make [] is {0}? (Also, dropping zero since it's really not used anywhere.)

srk/src/polynomialLattice.ml
nclskoh: That was a reflex :) Strengthened to require that the ideal generators of t1 lies in the ideal of t2; otherwise, multiplying a generator of t1 by a non-integer exits t2.

PR #61, opened 2023-06-02
AbstractLia to compute integer hull of LIA formulas

nclskoh:


srk/src/polyhedron.mli
```diff
@@ -101,6 +101,11 @@ val conical_hull : t -> t
 *)
 val integer_hull : ?decompose:bool -> [`GomoryChvatal | `Normaliz] -> t -> t
 
+(** [integer_hull_dd dim p] computes the integer hull of the double-description
+    polyhedron [p] in ambient dimension [dim].
+ *)
+val integer_hull_dd : int -> DD.closed DD.t -> DD.closed DD.t
```
zkincaid: Move to dD.ml.  Shouldn't need tot take ambient dimension as a parameter -- can be retrieved via ``Abstract0.dimension`` (only reald is used).

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+  let concretize p =
+    let p_formulas = implicant_of_polyhedral_constraints (DD.enum_constraints p) in
+    Syntax.mk_and context p_formulas
+
+  let abstract formula interp =
+    let implicant =
+      match Interpretation.select_implicant interp formula with
+      | Some implicant -> implicant
+      | None -> failwith "No implicant found" in
+    let (inequalities, _lattice_constraints) = constraints_of_implicant implicant in
```
zkincaid: Can't ignore lattice constraints

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+   - The row a^T Y - cx + b >= 0 (or = 0) that gives the lub, if one exists,
+     where c is positive
+   - The row a^T Y + cx + b >= 0 (or = 0) that gives the lub, if one exists,
+     where c is positive
+   - The other constraints that involve x
+   - The independent constraints that don't involve x
+   *)
+  type classified =
+    {
+      lub_row : (QQ.t * P.constraint_kind * V.t) option
```
zkincaid: Why compute both lub and glb?  Only need one or the other.  Can you select better terms by computing both?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+      )
+      {
+        lub_row = None
+      ; glb_row = None
+      ; others = BatEnum.empty ()
+      ; independent = BatEnum.empty ()
+      }
+      constraints
+
+  let _recession_cone_at m p =
```
zkincaid: ?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+    let hull = P.integer_hull_dd ambient_dimension p in
+    DD.project dimensions_to_eliminate hull
+
+  let pp fmt p =
+    Format.fprintf fmt "{ polyhedron : %a }"
+      (DD.pp (fun fmt d -> Format.fprintf fmt "%d" d)) p
+
+end
+
+module IntegerMbp : sig
```
zkincaid: Create a lattice polyhedron module for manipulating systems of linear inequalities & integrality constraints, and expose this functionality in that module.  Down the line, this should replace the model-based term selection inside quantifier.ml.

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
+open BatEnum.Infix
```
zkincaid, line 1: This should be in abstract.ml, and should replace the current AbstractRSY.  Currently interfaces do not quite match -- can you propose a common interface?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+  val bottom : t
+  val join : t -> t -> t
+  val concretize : t -> context Syntax.formula
+  val abstract : context Syntax.formula -> context Interpretation.interpretation -> t
+
+  val pp : Format.formatter -> t -> unit
+
+end
+
+module type Target = sig
```
zkincaid, line 26: Why should this be a module?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+             |> IntLattice.hermitize
+           in
+           (new_p, new_l)
+      ) (p, l) eliminate
+
+end
+
+module LatticePolyhedron : sig
+
+  val lattice_polyhedron_of : P.t -> IntLattice.t -> P.t
```
zkincaid: This can also be move to lattice polyhedron module.

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+end = struct
+
+  module T = Linear.MakeLinearMap(QQ)(Int)(V)(V)
+
+  let pp_linear_map fmt linear_map =
+    BatEnum.iter (fun (s, t) ->
+        Format.fprintf fmt "(%a, %a); " Linear.QQVector.pp s Linear.QQVector.pp t)
+      (T.enum linear_map)
+
+  let force_transform (forward, inverse, num_dimensions) p =
```
zkincaid: What is this doing?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+          | Some image ->
+             BatEnum.push q (kind, image);
+             (forward, inverse, num_dimensions)
+        )
+        (forward, inverse, num_dimensions)
+        (P.enum_constraints p)
+    in
+    (forward, inverse, num_dimensions, P.of_constraints q)
+
+  let transform map p =
```
zkincaid: What does "transform" mean?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+
+end
+
+module Abstract (A : AbstractDomain) : sig
+
+  val abstract : A.context Syntax.formula -> A.t
+
+end = struct
+
+  type t =
```
zkincaid: include init in abstract and eliminate this type.

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+         let next = { state with value = joined } in
+         begin match bound with
+         | Some b -> if n <= b then go (Some b) (n + 1) next
+                     else state
+         | None -> go bound (n + 1) next
+         end
+      | `Unsat ->
+         state
+      | `Unknown -> failwith "Can't get model"
+    in (go None 1 state).value
```
zkincaid: Eliminate "bound" -- looks like it is for debugging purposes.

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+           in
+           let difference = QQ.sub (m dim) (Linear.evaluate_affine m glb_term) in
+           let residue = QQ.modulo difference (QQ.of_zz divisor) in
+           let solution = V.add_term residue Linear.const_dim glb_term in
+           logf ~level:`trace "glb value %a <= point %a, difference %a, divisor %a, residue %a@."
+             QQ.pp (Linear.evaluate_affine m glb_term) QQ.pp (m dim)
+             QQ.pp difference QQ.pp (QQ.of_zz divisor) QQ.pp residue;
+           logf ~level:`trace "glb term %a@." V.pp glb_term;
+           logf ~level:`trace "selected term %a, <= %a:1@." V.pp solution pp_dim dim;
+           let () = match classified.lub_row with
```
zkincaid: Why not just keep lub_row/glb_row in "others"?

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+           in
+           let new_p =
+             classified.others
+             /@ (fun (kind, a) ->
+               (kind, substitute_term solution dim a))
+             |> BatEnum.append classified.independent
+             |> P.of_constraints in
+           let new_l =
+             List.map (substitute_term solution dim) (IntLattice.basis l)
+             |> List.cons solution
```
zkincaid: "solution" necessary?  I would think that that you would need to add IsInt(x) for each int-sorted symbol x at the outset.

srk/src/abstractLia.ml
nclskoh: Do you mean to move the whole of abstractLia into abstract?

srk/src/abstractLia.ml
nclskoh: Yeah this is shoddy, removed it and improved the presentation.

srk/src/abstractLia.ml
nclskoh: yes that's done, so cons-ing "solution" is unnecessary; removed.

srk/src/abstractLia.ml
nclskoh: Moved

srk/src/abstractLia.ml
nclskoh: Moved

srk/src/abstractLia.ml
nclskoh: Yup, done

srk/src/abstractLia.ml
nclskoh: done

srk/src/polyhedron.mli
nclskoh: I've moved this and Gomory-Chvatal to DD. But there is now a circular dependency between DD and Cone, because of Hilbert basis computation, which calls out to Normaliz.

srk/src/abstractLia.ml
nclskoh: destruct_atom in Interpretation.ml currently translates IsInt to a mod term, so a lattice constraint actually never shows up here. I think we should change it to let IsInt pass through, but this affects how the wedge domain is handling the implicant.

srk/src/abstractLia.ml
nclskoh: This is wrong; it's supposed to extend the linear map to dimensions outside of the lattice so that the polyhedron can be mapped to the new space, but now the integer hull of the polyhedron in that space is not we want. I have to think about what we should do for real-integer lattices, but at present, I've changed the code to assume a full-rank lattice and just apply the standardization map preserving 1.

srk/src/abstractLia.ml
nclskoh: It transforms the polyhedron by applying the map to its dual constraints. Is it the name that's confusing?

srk/src/abstractLia.ml
nclskoh: This looks like it should be a record instead, but it falls out of the current design that is module- and functor-centric; in particular, ImplicantConstraintConversion organizes functions that have to do with conversion between implicants and polyhedra over the underlying arithmetic terms. I've tried doing things in terms of records, but module types seem to make things look neater.

srk/src/abstractLia.ml
nclskoh: Now computes lattice hull via LatticePolyhedron

srk/src/dD.ml
```diff
@@ -175,6 +177,155 @@ let minimal_faces polyhedron : (V.t * ((constraint_kind * V.t) list)) list =
...
+  let sparsify ctx v =
+    BatList.fold_lefti (fun vec i a ->
+        V.set (D.dim_of_int ctx i) (QQ.of_zz (ZZ.of_mpz a)) vec)
+      V.zero
+      v
+
+  (* TODO: This is basically Cone's [hilbert_basis] function, but there is
+     a dependency cycle between Cone and DD.
+   *)
+  let hilbert_basis lines rays =
```
zkincaid, line 201: Leave the dense variant here and put the sparse one in cone.ml?

srk/src/latticePolyhedron.ml
```diff
@@ -0,0 +1,170 @@
+module V = Linear.QQVector
+module P = Polyhedron
+module L = IntLattice
+module T = Linear.MakeLinearMap(QQ)(Int)(V)(V)
+
+include Log.Make(struct let name = "srk.latticePolyhedron" end)
+
+let map_polyhedron map p =
```
zkincaid, line 8: Geometrically, what does this do?

srk/src/latticePolyhedron.ml
```diff
@@ -0,0 +1,170 @@
...
+
+  let pp_dim = fun fmt d -> Format.fprintf fmt "%d" d
+
+  let local_project_cooper m ~eliminate (p, l) =
+    BatList.fold_left (fun (p, l) dim ->
+        let (classified, has_upper_bound) =
+          classify_constraints m dim (P.enum_constraints p) in
+        if not (has_upper_bound) || classified.glb_row = None then
+          ( P.of_constraints classified.irrelevant
+          , IntLattice.project (fun dim -> not (BatList.mem dim eliminate)) l )
```
zkincaid: Should only project dim out of the lattice rather than all dimensions

srk/src/latticePolyhedron.ml
nclskoh: This applies the inverse of the adjoint of [map] to the polyhedron [p].

srk/src/polyhedron.mli
zkincaid: Can you move the dense variant of Hilbert basis computation to DD and the sparse version in cone.ml?

srk/src/abstractLia.ml
zkincaid: Yes, assuming that it can replace AbstractRSY.

srk/src/abstractLia.ml
zkincaid: Could expose multiple variants of ```destruct_atom```, one for each theory.  ```destruct_lia_atom''' could could retain IsInt.

srk/src/abstractLia.ml
```diff
@@ -0,0 +1,626 @@
...
+      )
+      (P.enum_constraints p);
+    P.of_constraints q
+
+  let pp_dim = fun fmt d -> Format.fprintf fmt "%d" d
+
+  let lattice_polyhedron_of p l =
+    let basis = IntLattice.basis l in
+    let (forward_l, inverse_l, num_dimensions_l) =
+      List.fold_left (fun (forward, inverse, index) v ->
```
zkincaid: fold_lefti

srk/src/abstractLia.ml
zkincaid: Yes, a more descriptive name would be helpful.  Does this have a geometric interpretation?

srk/src/latticePolyhedron.ml
zkincaid: I don't follow -- does "applies" mean "compute the image of", "compute the inverse image of", something else?  (If a polyhedron lives in Q^n, then its cone of valid inequalities live in Q^{n}* x Q, so [map] must be a linear map Q^{n}* x Q -> Q^{m}* x Q, and I don't see how this should be interpreted.  A possible explanation is what's happening is that we're computing the image of a cone under a map, and that cone happens to be the cone of valid inequalities of a polyhedron, but we shouldn't attempt make sense of this operation on the polyhedra themselves)

nclskoh: [map] is an invertible linear map from Q^{n} * Q -> Q^{n} * Q that preserves the constant dimension, and is applied to the cone of valid inequalities of the polyhedron P. When we compute the image of this cone under [map] and interpret it as constraints of a polyhedron Q, i.e., the result of the function, Q = (adjoint(map))^{-1}(P).

This interpretation depends on [map] preserving 1 and being an endomorphism whose image contains P (I think); it certainly works when [map] is invertible and preserves 1.

srk/src/abstractLia.ml
nclskoh: good point

srk/src/abstractLia.ml
nclskoh: This is now under [map_polyhedron] and the discussion there. I guess the name is still somewhat up in the air.

srk/src/dD.ml
nclskoh: The one here is also handling sparse vectors because that's what we get from within DD. Do you mean to export this in DD.mli with the signature [lines:QQVector.t list -> rays:QQVector.t list -> QQVector.t list] and have the one in cone.ml have its current signature [t -> QQVector.t list] call out to this one?

zkincaid: Within DD, vectors are dense (but yes, represented by the type QQVector.t).  I think the interface (lines:QQVector.t list -> rays:QQVector.t list -> QQVector.t list) or (rays:QQVector.t list -> QQVector.t list) in DD.ml is reasonable.

PR #62, opened 2023-07-27
Merge Multiplicative

cyphertjohn:
Contains at least the following

 - Removes OCRS
 - Adds F4 interface
 - Adds refinement
 - Adds number field operations
 - Adds finding algebraic relations of solvable polynomial maps
 - Adds interface and functionality for polynomials over a field, which forms a Euclidean domain. I.e. gcd, square free factor, etc.


duet/cra.ml
```diff
@@ -138,6 +141,74 @@ module K = struct
...
+
+  (*
+  let mul x y = Log.time "refine" mul x y
+  let add x y = Log.time "refine" add x y
+  *)
+
+  module CRARefinement = Refinement.DomainRefinement
+      (struct
+        include Tr
+        let equal a _ = ((LirrSolver.is_sat srk (guard a)) == `Unsat)
```
zkincaid: This line looks suspicious.  Does DomainRefinement really just need an is_zero function exposed rather than equality?

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+    equivalent univariate polynomial.
+    @raise Failure if input is not a univariate polynomial.*)
+val make_univariate : QQXs.t -> QQX.t
+
+(** Let K1 = Q[v0,v1]/p and K2=Q[v0,v1]/q be number fields. Let K1 have degree d1 and K2 have degree d2.
+    [prim, v0_in_prim, v1_in_prim] = [primitive_elem (d1*d2) p q v0 v1] 
+    is such that the number field K3 = Q[x]/prim of degree d1*d2 such that K3 contains K1 and K2. 
+    [v0_in_prim] and [v1_in_prim] give the representations of v0 and v1 in K3 respectively. For
+    the method to work p needs to be irreducible in K2 and q needs to be irreducible in K1.*)
+val primitive_elem : int -> QQXs.t -> QQXs.t -> Monomial.dim -> Monomial.dim -> QQX.t * QQX.t * QQX.t
```
zkincaid, line 17: I would expect the number fields to take the form Q[x]/p for some p.  What's the second variable?

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+
+  val sub : elem -> elem -> elem
+
+  val inverse : elem -> elem
+
+  val exp : elem -> int -> elem
+
+  val equal : elem -> elem -> bool
+
+  val compare : elem -> elem -> int
```
zkincaid: Presumably this is an arbitrary linear order on the number field?  E.g., we shouldn't rely on comparison to be consistent with the usual one for algebraic reals?

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+                                                            \[\[f_11; ...; f_1k\];
+                                                            \[f_21; ...; f_2k\];
+                                                            \[...\]
+                                                            \[f_l1; ...; f_lk\]\]
+        for the set {n_1, ..., n_k : gamma_1^n_1...gamma_k^n_k = 1.} *)
+  val find_relations : elem list -> int list list
+
+end
+
+(** Given a minimal polynomial, construct a number field. Giving a
```
zkincaid: Does minimal just mean irreducible?  (Or should I interpret this as "construct Q(a) from the minimal polynomial of a"?)

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+
+      (** Convert a rational polynomial to a polynomial in the number field*)
+      val lift : QQX.t -> t
+
+      (** Convert a polynomial in the number field to a multivariate polynomial.
+          The field variable is 0, and the polynomial variable is 1.*)
+      val de_lift : t -> QQXs.t
+
+      (** Factors a square free polynomial in the number field.*)
+      val factor_square_free_poly : t -> (scalar * ((t * int) list))
```
zkincaid: Assuming that the (t * int) list should be interpreted as factors and their corresponding powers.  If the input is square-free, why do we need the powers?

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+  val deg : int
+
+  (** int_poly is a monic integer polynomial that defines the same field as A.min_poly. *)
+  val int_poly : QQX.t
+
+  (** The type of elements of the number field. *)
+  type elem
+
+  (** Converts a univariate polynomial to an element of the number field.*)
+  val make_elem : QQX.t -> elem
```
zkincaid: I'm not sure how to use make_elem as a client of this module.  For example, if I construct a number field from the min polynomial of some a, then does the polynomial "x" correspond to a?  (That sounds fine if that's what it is.  My source of confusion is that the min polynomial of a does not uniquely define a.  I suspect that the resolution is that it just doesn't matter: there's an isomorphism Q[x]/p -> Q(a) for each root a of p, and so we effectively just need to pick one and be consistent).

zkincaid: (Would it be a useful operation to be able to get a list of all of the roots of the defining polynomial of the number field?  I guess that your interface already supports this by factoring the polynomial and calling extract_root_from_linear on each factor)

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
+exception Divide_by_zero
+
+module type RationalFunc = sig
+
+  type num
+
+  type den
```
zkincaid, line 7: Is it conceivable that num != den?  Is RationalFunc really just rational functions, or can we think of it more generally as a field of fractions?

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
...
+  val mul : t -> t -> t
+
+  val int_mul : int -> t -> t
+
+  val pp : Format.formatter -> t -> unit
+
+  (**Computes partial fractional decomposition. Let [den_pow] = (g_1, e_1), ..., (g_k, e_k).
+  [partial_fraction num den_pow] computes (n_1, (g_1, e_1)), ..., (n_k, (g_k, e_k))
+  such that num/((g_1, e_1)...(g_k, e_k)) = (n_1/(g_1, e_1)) + ... + (n_k/(g_k, e_k))*)
+  val partial_fraction : num -> (den * int) list -> (num * (den * int)) list
```
zkincaid, line 41: Do denominators need to be irreducible (or just coprime, or doesn't matter)?

srk/src/rational.mli
zkincaid: I see MakeRatDiff.  So then should we think of this as a localization of a ring?

srk/src/expPolynomial.ml
```diff
@@ -2,46 +2,85 @@ open BatPervasives
 open Syntax
 
 module QQX = Polynomial.QQX
-module E = Ring.RingMap(QQ)(QQX)
 module QQMatrix = Linear.QQMatrix
 
-type t = E.t
+module MakeEP 
+  (B : sig
```
zkincaid, line 8: Do we make use of exponential polynomials with B != C?

srk/src/polynomial.ml
```diff
@@ -1249,35 +1334,132 @@ module Rewrite = struct
...
+    ) rat_fmon
+
+  let convert_from_faugere (vs : Monomial.dim list) (p : fpoly) : QQXs.t =
+    let convert_fmon_to_mon (c, fm) = 
+      Q.of_bigint c, Monomial.of_enum (BatList.enum (List.mapi (fun i deg -> List.nth vs i, deg) fm)) in
+    QQXs.of_list (List.map convert_fmon_to_mon p)
+
+  (* Is this right? *)
+  let get_mon_order (blk1 : Monomial.dim list) (blk2 : Monomial.dim list) a b = 
+    let ablk1, ablk2 = Monomial.split_block (fun v -> if (List.mem v blk1) then true else false) a in
```
zkincaid: `if (List.mem v blk1) then true else false` -> `List.mem v blk1`

srk/src/polynomial.ml
```diff
@@ -66,28 +67,109 @@ module MakeUnivariate(R : Algebra.Ring) = struct
...
+      let rec aux q r =
+        let rd = order r in
+        if rd = 0 || rd < d then (q, r)
+        else
+          let s = (of_term (F.mul (coeff rd r) (F.inverse c)) (rd - d)) in
+          aux (add q s) (sub r (mul s b))
+        in
+      aux zero a
+  
+  let ex_euc a b =
```
zkincaid: Zarith / GMP call this "gcdext"

srk/src/polynomial.ml
```diff
@@ -1294,16 +1476,192 @@ module Ideal = struct
...
-  let equal = equal
 
-  let make generators =
-    grobner_basis (mk_rewrite order generators)
+  module IS = SrkUtil.Int.Set
+
+  type fgb_t = 
+  {
+    basis : FGb.fpoly list
+  ; vs : IS.t
```
zkincaid, line 1487: vs is the set of variables over which the polynomials are expressed?

srk/src/polynomial.ml
```diff
@@ -1294,16 +1476,192 @@ module Ideal = struct
...
+    let vs = List.fold_left (fun acc p -> IS.union acc (QQXs.dimensions p)) IS.empty generators in
+    let fpolys = List.map (to_fmon vs) generators in
+    let basis = FGb.grobner_basis_fmon (BatList.of_enum (IS.enum vs)) [] fpolys in
+    {basis; vs}
+
+  let intersect_f j k =
+    let vs = IS.union j.vs k.vs in
+    let j = generators_f j in
+    let k = generators_f k in
+    let t =
```
zkincaid: t is 1 + IS.max_elt vs?

srk/src/transition.ml
```diff
@@ -195,7 +195,7 @@ struct
   let star tr =
     let (module D) = !domain in
     let tf = to_transition_formula tr in
-    logf "Approximating transitive closure:@.  @[<v 0>%a@]" pp tr;
+    logf ~level:`warn "Approximating transitive closure:@.  @[<v 0>%a@]" pp tr;
```
zkincaid, line 198: Why warn?

srk/src/numberField.mli
```diff
@@ -0,0 +1,214 @@
...
+end
+
+(** Given a minimal polynomial, construct a number field. Giving a
+    polynomial that is not irreducible in Q[x] gives undefined behavior.*)
+module MakeNF (A : sig val min_poly : QQX.t end) : NF
+
+(** [min_poly, roots] = [splitting_field p] is such that Q[x]/min_poly is the
+    splitting field of [p]. [roots] give the representation of the roots along with
+    their multiplicity of [p] in the splitting field.*)
+val splitting_field : QQX.t -> QQX.t * ((QQX.t * int) list)
```
zkincaid: Could make this a functor?

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+end
+
+
+
+
+
+
+module MakeNF (A : sig val min_poly : QQX.t end) = struct
+
+  type elem = QQX.t (* elements of the field are polynomials of degree < the degree of the field*)
```
zkincaid: Rename to elem -> t (so e.g., we can create a set of NF elements with Set.Make(NF) without having to rename elem -> t).

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+      let lc = QQX.coeff (QQX.order min) min in
+      QQX.map (fun _ c -> QQ.div c lc) min
+
+
+  (* The lcm of the denominators of min_poly*)
+  let min_poly_den_lcm = 
+    QQX.fold (
+      fun _ c l ->
+        ZZ.lcm (QQ.denominator c) l
+    ) A.min_poly (QQ.numerator (QQX.coeff (QQX.order A.min_poly) A.min_poly))
```
zkincaid, line 149: `(QQ.numerator (QQX.coeff (QQX.order A.min_poly) A.min_poly))` -> `QQ.one`?

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+      ~pp_sep:(fun formatter () -> Format.fprintf formatter "@ + ")
+      pp_monomial
+      formatter
+      (QQX.enum p)
+
+  
+
+  let int_mul i x = QQX.scalar_mul (QQ.of_int i) x
+
+  module E = struct let one = one let mul = mul let negate = negate let exp = exp let pp = pp end
```
zkincaid: extend E with additional operations so that line 245 can become include MakeEuclidean(E)

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+      let m = Array.make_matrix p_deg (p_deg+1) QQ.zero in (*m is a matrix whose columns consist of the powers of el*)
+      for i = 0 to p_deg do
+        let r = snd (QQX.qr (QQX.exp el i) p) in
+        BatEnum.iter (
+          fun (c, d) -> 
+            m.(d).(i) <- c
+        ) (QQX.enum r)
+      done;
+      (*In general I don't think null is guarenteed to be reduced with respect to powers of el. However,
+         I think it is do to how nullspace is implemented. *)
```
zkincaid: Meaning: we need the polynomial of minimal degree that belongs to the nullspace, and we're relying on the fact that it must belong to the basis?

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+    QQX.add
+  
+  let sub = 
+    QQX.sub
+
+  let inverse a = 
+    let (_, u, _) = QQX.ex_euc a int_poly in
+    u
+
+  let exp a i =
```
zkincaid, line 193: `Srk.exp mul QQX.one a i` might be faster than `reduce (QQX.exp a i)` -- keeps the polynomial small by reducing with each mul.

srk/src/numberField.ml
```diff
@@ -0,0 +1,994 @@
...
+
+
+
+
+
+module MakeNF (A : sig val min_poly : QQX.t end) = struct
+
+  type elem = QQX.t (* elements of the field are polynomials of degree < the degree of the field*)
+
+  let compute_min_poly_p el p =
```
zkincaid: Can this be moved outside the MakeNF module?

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+  let get_normalized a = 
+    match !a with
+    | Norm (n, d) -> (n, d)
+    | Unnorm (n, d) ->
+      if D.equal n D.zero then 
+        (a := Norm (D.zero, D.one);
+        (D.zero, D.one))
+      else
+        let gcd = gcd n d in
+        let res = div_factor n gcd, div_factor d gcd in (* is this normal? could have const factor*)
```
zkincaid: Spec for ex_eux says gcd is monic, so I don't think it's necessarily normal.  Need to multiply gcd by `QQX.coeff (QQX.order d) d`.

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+
+  let negate a =
+    match !a with
+    | Unnorm (p, q) -> ref (Unnorm (D.negate p, q))
+    | Norm (p, q) -> ref (Norm (D.negate p, q))
+
+  let inverse a = 
+    match !a with
+    | Unnorm (p, q) when not (D.equal q D.zero) -> ref (Unnorm (q, p))
+    | Norm (p, q) when not (D.equal q D.zero) -> ref (Norm (q, p))
```
zkincaid: (Assuming (equal x y) should imply (equal_syn x y) when x and y are normal, need to divide p and q by `QQX.coeff (QQX.order p) p`

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+end)
+(CX : sig
+  include Univariate with type scalar = C.t
+  val pp : Format.formatter -> t -> unit
+end) = struct
+  module E = ExpPolynomial.MakeEP(B)(C)(CX)
+
+  type t = {
+    ep : E.t;
+    heaviside : C.t IM.t
```
zkincaid, line 206: Might be slightly simpler  with IM = Ring.MakeVector(C)

zkincaid: For example, `zero` and `of_heavy 1 C.zero` should be equal(?), but are not.  Vectors are zero-suppressed, so `of_heavy x C.zero` is the same as the empty map.

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+
+  type den = D.t
+
+  let gcd a b = 
+    let g, _, _ = D.ex_euc a b in
+    g
+
+  let div_factor a factor = fst (D.qr a factor)
+
+  type pre_t = Norm of (D.t * D.t) | Unnorm of (D.t * D.t)
```
zkincaid, line 44: I assume "Norm" means numerator and denominator are coprime, and denominator is monic?

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+        List.fold_left (fun acc t -> EP.add acc (translate_term t)) sum terms
+      in
+      BatEnum.fold translate_and_add ep_with_e_and_heavy iifs_rat_pow
+    in
+    let eps_nf = Array.map translate_ep eps in
+    EP.set_rec_sols eps_nf;
+    let ep = (module EP : ExpPolyNF) in
+    ep
+  
+  let exp a i =
```
zkincaid, line 1236: SrkUtil.exp should cover this

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+      BatEnum.fold (fun acc ep -> RS.add acc (translate_iif ep)) ep_rs_heavy (enum_iif p_ep)
+    in
+    let rec aux blocks offset = 
+      match blocks with
+      | [] -> ()
+      | (blk : block) :: blks ->
+        let blk_size = Array.length (blk.blk_add) in
+        let blk_add = Array.map translate_blk_add blk.blk_add in
+        let init = Array.sub initial offset blk_size in
+          (*match initial with
```
zkincaid: ?

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
...
+  val of_polynomial : ConstRingX.t -> t
+
+  (**Exponential bases are number field elements*)
+  val of_exponential : NF.elem -> t
+
+  val of_exponential_poly : NF.elem -> ConstRingX.t -> t
+
+  (**One way to handle 0 eigenvalues is to use heavyside functions. Essentially iverson brackets multiplied by constants
+      of_heavy i c represents the term \[n>=i\] * c.*)
+  val of_heavy : int -> ConstRing.t -> t
```
zkincaid, line 103: (Heaviside?)
For example, if we have x_n = x_{n-2}, then the solution is x_0 with heaviside [ 1 -> -x_0 + x_1, 2 -> -x_1 ]?

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
...
+
+  val scalar_mul : ConstRing.t -> t -> t
+  val add : t -> t -> t
+  val negate : t -> t
+  val equal : t -> t -> bool
```
zkincaid, line 108: Does equality mean "equality as functions" or "syntactic equality"?  E.g., p is the same as (heavy 0 p)

srk/src/rational.ml
```diff
@@ -0,0 +1,1299 @@
...
+    let a_init = N.fold (build_init (a_comp.(0))) an (Array.make (Array.length (a_comp.(0))) ConstRing.zero) in
+    let b_init = N.fold (build_init (b_comp.(0))) bn (Array.make (Array.length (b_comp.(0))) ConstRing.zero) in
+    let transform = kronecker QQ.mul a_comp b_comp in
+    let init_m = kronecker ConstRing.mul [|a_init|] [|b_init|] in
+    let init = init_m.(0) in
+    let sol = solve_mat_rec_rs transform init (Array.make (Array.length init) zero) in
+    let n, p = Array.length a_comp, Array.length b_comp in
+    sol.(n*p - p - 2) (* Need to check this*)
+
+(*let had_exp rs e =
```
zkincaid, line 887: ?

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
...
+  (** Enumerate the exponential part.*)
+  val enum_ep : t -> (ConstRingX.t * NF.elem) BatEnum.t
+
+  (** Enumerate the heavysides*)
+  val enum_heavy : t -> (int * ConstRing.t) BatEnum.t
+
+  val pp : Format.formatter -> t -> unit
+  
+  (** This module contains recurrence solutions.*)
+  val get_rec_sols : unit -> t array
```
zkincaid, line 120: Slightly odd that recurrence solutions are inside the module.  Assuming that this is because we want RatEP.to_nf to return an ExpPolyNF A *and* an array of A.t's, but this is a pain in OCaml's type system.  I think your solution is ok, but just wanted to mention that GADTs are one way around this:

```
type soln = Soln : (module ExpPolyNF with type t = 'a) * 'a array -> soln
```
then RatEP.to_nf can return a soln.

srk/src/solvablePolynomial.ml
```diff
@@ -1040,10 +937,129 @@ let extract_constant_symbols srk tr_symbols wedge =
...
-  let open Type_def in
 
+let term_of_ratep srk loop_counter pre_term_of_id ep =
+  let module CX = Rational.ConstRingX in
+  let open Rational.RatEP in
+  let translate_const_ring c = 
+    let translate_mon m = 
+      BatEnum.fold (
+        fun mul (dim, pow) ->
+          (mk_pow srk (pre_term_of_id dim) pow) :: mul (*mk_int or mk_real?*)
```
zkincaid: I don't understand the comment

zkincaid: (or the  same one below)

srk/src/solvablePolynomial.ml
```diff
@@ -1040,10 +937,129 @@ let extract_constant_symbols srk tr_symbols wedge =
...
+        term :: acc
+    ) p [] in
+    mk_add srk add_l
+  in
+  let translate_ep (p, b) = 
+    mk_mul srk [translate_poly p; Nonlinear.mk_pow srk (mk_real srk b) loop_counter]
+  in
+  let translate_iif ((den, shift), c) = 
+    let func = QQX.show den in (*Converting the iif to a string*)
+    let arg = if shift = 0 then loop_counter else mk_add srk [loop_counter; mk_real srk (QQ.of_int shift)] in (*mk_real or mk_int?*)
```
zkincaid: `mk_int srk shift` is the same as `mk_real srk (QQ.of_int shift)`

srk/src/solvablePolynomial.ml
```diff
@@ -1955,25 +1947,125 @@ end
...
+      Format.fprintf formatter "x_%d'" (i - offset)
+
+
+  let exp_ti it = 
+    let it_offset = it.ideal.dim in
+    logf  "Exponentiating : %a" (TransitionIdeal.pp (pp_dim it_offset)) it.ideal;
+    if I.generators (it.ideal.ideal) = [] then
+      TransitionIdeal.make it_offset it.ideal.ideal
+    else if QQXs.is_zero (I.reduce it.ideal.ideal QQXs.one) then
+      TransitionIdeal.make it_offset it.ideal.ideal
```
zkincaid: Should be the identity

srk/src/solvablePolynomial.ml
```diff
@@ -1955,25 +1947,125 @@ end
...
+        let pp_cl f = Array.iteri (fun i cl -> Format.fprintf f "cl.(%d) = %a@." i Rational.RatEP.pp cl) in
+        logf  "%a" pp_cl cf;
+        let sp_map_offset = Array.length cf in
+        let module EP = (val Log.time "Splitting Field" Rational.RatEP.to_nf cf) in
+
+        let zero_eig_transient, zero_eigen_stab, rels = Log.time "Algebraic Relations" EP.long_run_algebraic_relations () in
+        logf  "Alg Relations: %a" (Id.pp (pp_dim it_offset)) (Id.make rels);
+        
+        let cl = 
+          if (List.length inv_seq_id) + 1 >= zero_eigen_stab then
```
zkincaid: Not sure that I'm reading correctly, but my understanding of this is that you want the transient length to be the max of the iteration at which the domain of ti converges and the size of the largest jordan block of 0.  That seems plausible, but I'm not sure: the solvability witness approximates the behavior of ti + dom*(ti), but not (necessarily) ti.

Consider x' = x /\ y' = x /\ y = 0 /\ z_1' = xz_1 /\ z_2' = z_1 /\ ... /\ z_n' = z_{n-1}
This transition ideal has invariant domain <x, y> (converging @ iteration 2) and solvability witness
f(x,y,z_1,...,z_n) = (0, 0, 0, z_1, ... , z_{n-1} )
(with a Jordan block of size n).  My thought is that we need a transient length of n+2 rather than max(n,2).

srk/src/expPolynomial.ml
cyphertjohn: Yes. In rational.ml C is a const ring of coefficients. I.e. multivariate polynomials over a number field and variables are symbolic constants. Whereas B are elements of a number field. There they need to be separate, because I can compute algebraic relations among exponentials over a number field, but not w.r.t exponentials where the bases are multivariate polynomials.

srk/src/numberField.ml
cyphertjohn: Yes, that is the assumption. Take the example of Q[x]/p(x) where p(x) = x^4 - 2x^2 + 9. Then find the minimal polynomial of i = (x^3 + x)/6. i is really the imaginary unit. Taking powers of i and then computing the nullspace gives, the following possible basis. (1, 0, 1, 0, 0), (0, 1, 0, 1, 0), (-1, 0, 0, 0, 1). Corresponding to i^2 = -1, i^3 = -i, and i^4 = 1. The first basis element is minimal because the basis has been oriented to put higher powers to the right. If we use rref to compute the null space we can be sure that the minimal vector will be in the basis. nullspace uses rref so I think the assumption holds. However, this property is not stated for the interface of nullspace, so if the implementation of nullspace changes the assumption might not hold. I used nullspace because I couldn't find an accessible interface to compute rref.

srk/src/numberField.ml
cyphertjohn: No, A.min_poly is the polynomial given to construct the NF. It is not assumed to be monic for the implementation. On instatiation of the functor, the implementation computes a monic integer polynomial accessible as int_poly such that A.min_poly and int_poly generate the same number field. int_poly is the the minimal polynomial that will be used for the rest of the calculations.

srk/src/numberField.mli
cyphertjohn: This is messy, and I couldn't quite come up with a good interface. If all you want is to compute a primitive element for two number fields Q[x]/p and Q[y]/q, then variables and such don't need to be stated. However, for factoring and computing a splitting field we want to compute a "primitive element" for something like Q[x,y]/<p(x),q(x, y)>. In this case the y roots of q(x,y) depend on x. So we need both variables. You probably could simplify the interface to only computing a primitive element for fields of the form Q[x]/p(x). But this would be extra expensive for the cases that are needed for factoring and computing a splitting field.

srk/src/numberField.mli
cyphertjohn: See earlier comment for a distinction between A.min_poly and int_poly. Also, I'm not sure I exactly understand what you're saying. First, remember that the minimal poly does *not* factor in the number field. You will get at least two factors, but not more than that. So at this point we do not know the factors of either int_poly or A.min_poly. To figure them out we need to compute a splitting field first, then factor. R.e to your first comment. Let alpha be the root of  A.min_poly corresponding to x. For the implementation of the order which comes later, it is convenient to have a algebraic integer in the number field. At instantiation we compute such an integer as follows. If A.min_poly is a monic integer polynomial, then we take the algebraic integer alpha, and int_poly = A.min_poly. If A.min_poly is not a monic integer polynomial we compute the lcm of the dens of A.min_poly. Then beta = lcm*alpha is an algebraic integer. int_poly is then the minimal polynomial of beta. Then internally elements of the number field are represented as univariate polynomials modulo the minimal polynomial of beta, say q(x). The purpose of make_elem is that externally, the user supplied A.min_poly <> q(x). And thus assumes the field is Q[x]/A.min_poly. But this is not the case, it's Q[x]/q(x). Make_elem recognizes the isomorphism and performs the conversion. E.x. if A.min_poly = x^2+1, then make_elem x = x. I.e. x is represented as x. However, if A.min_poly = x^2 + 1/4, then int_poly = x^2 + 4. Then make_elem x = 1/4 x. Externally, x is either -i/2 or i/2 and for any computations of NF it doesn't matter which you choose. Internally, x corresponds to either -2i or 2i. Again it doesn't matter which one you use.

srk/src/numberField.mli
cyphertjohn: Correct. No mathematical assumptions should be made on compare.

srk/src/polynomial.ml
cyphertjohn: Yes

srk/src/rational.ml
cyphertjohn: Should be. I may have made a mistake somewhere though.

srk/src/rational.ml
cyphertjohn: Recurrences can be represented by exponential polynomials or rational sequences. Before, I was representing them by rational sequences when solving a solvable polynomial map. In that case you need to be able to multiply two sequences, or exponentiate a sequence with hadamard multiplication. In practice this turned out to be expensive. So I instead used exponential polynomials to represent rational sequences. They have their own multiplication as exponentiation which is in practice cheaper. So I switched and this became deprecated. I can remove if needed.

srk/src/rational.mli
cyphertjohn: I'm not as familiar, but I think that's correct.

srk/src/rational.mli
cyphertjohn: I don't think it matters.

srk/src/rational.mli
cyphertjohn: syntactic equality

srk/src/transition.ml
cyphertjohn: Info prints too much stuff. I.e. all the lirr logging. I wanted a level to just see the analysis of loops. It was easier to understand what's going on when running on examples. I can change it back.

srk/src/numberField.mli
cyphertjohn: Changed to irreducible

srk/src/numberField.mli
cyphertjohn: Modified to only return factors.

srk/src/solvablePolynomial.ml
cyphertjohn: I wasn't sure what the difference was at the time. I removed the comment

srk/src/solvablePolynomial.ml
cyphertjohn: changed to mk_int

srk/src/polynomial.ml
cyphertjohn: Fixed

srk/src/polynomial.ml
cyphertjohn: ex_euc -> gcdext

srk/src/numberField.ml
cyphertjohn: Fixed

srk/src/numberField.ml
cyphertjohn: elem -> t

srk/src/rational.ml
cyphertjohn: Removed

srk/src/numberField.mli
cyphertjohn: I'm not sure about making it a functor. But I made it return a number field module now. I.e. splitting_field : QQX.t -> (module NF with type t = 'a) * (('a * int) list). Not exactly this, but using a GADT.

srk/src/rational.mli
cyphertjohn: I left it as is

srk/src/numberField.ml
cyphertjohn: Now uses Srk.exp mul QQX.one a i

srk/src/numberField.ml
cyphertjohn: Done

srk/src/rational.ml
cyphertjohn: Fixed

srk/src/rational.ml
cyphertjohn: Fixed

srk/src/rational.ml
cyphertjohn: Changed

srk/src/polynomial.ml
cyphertjohn: Changed

duet/cra.ml
cyphertjohn: Changed to is_zero

srk/src/rational.mli
cyphertjohn: I'm not sure I understand your example. If x_n = x_{n-2} then x_n = (1/2*x_0 + (-1/2)*x_1)(-1)^x + 1/2*x_0 + 1/2*x_1

srk/src/rational.ml
cyphertjohn: I've changed of_heavy to check for the case of zero. It doesn't fully address your comment of whether Ring.MakeVector(C) would be better. I will say that the only way to get a "zero" heavy is by of_heavy 1 C.zero. In all other cases it should be impossible for a heavy to be zero.

srk/src/numberField.mli
zkincaid: So should I think of Q[x]/p as a representation of  the number field Q(a) for some a, and Q[x,y]/<p(x),q(x,y)> as a representation of Q(a,b) where b is a root of q(x,y), considered as a univariate polynomial over Q(a)?

srk/src/numberField.ml
zkincaid: One way to solve this without this assumption:
`module M = Linear.MakeLinearMap(QQ)(SrkUtil.Int)(QQX)(QQX)`
creates a module M representing linear maps Q[x] -> Q[x].  You can construct a linear map by adding bindings p^0 -> 1, p^1 -> x, p^2 -> x^2 to a map m ... until eventually `M.apply m p^n` is defined (returns a polynomial rather than None).  The returned polynomial is the minimal polynomial.

srk/src/numberField.ml
zkincaid: Why is the numerator of the leading coefficient part of the lcm of the denominators of min_poly?

srk/src/numberField.mli
zkincaid: My basic confusion is that I wouldn't expect the client to know that internally Q(a) is represented by QQ[x]/p (where p is the supplied minimal polynomial of a or some other polynomial). But I think it's ok if we do want to expose that representation.

srk/src/rational.mli
zkincaid: Oops --  I guess the recurrence I was looking for was x_{n+2} = 0 (so the sequence is [x_0,x_1,0,0,...])

srk/src/rational.ml
zkincaid: `scalar_mul C.zero (of_heavy 1 C.one)`?

srk/src/numberField.ml
cyphertjohn: If A.min_poly = 4x^2 + 1, the field is Q(i/2). But i/2 is not an algebraic integer. We need the lcm of the monic polynomial that defines Q(i/2). I.e. x^2 + 1/4. The lcm of the dens of x^2 + 1/4 is 4, which is the lc of A.min_poly.

srk/src/numberField.mli
cyphertjohn: Yes. This function is a little messy. We could keep it internal to the .ml file if you think the interface is too complicated.

srk/src/rational.mli
cyphertjohn: Such a solution is represented as x_0 = x_0 + (-x_0 + x_1)[x>=1] + (-x_1)[x>=2]. The heavisides are terms within a sum.

srk/src/numberField.mli
cyphertjohn: For computing algebraic relations of exponential polynomials in rational.ml I need to get ahold of some representation of field elements as polynomials. This is because we get algebraic relations over only rational numbers by creating an ideal I involving polynomials p1(x), ..., pk(x) \in Q[x]/q for some q. p1, ..., pk represent number field elements we get purely rational relations by creating <I, p1(x), ..., pk(x), q(x)> and eliminating x. There is a question of whether the user should be able to supply number field elements as polynomials. This is the purpose of make_elem. I use make_elem only in test_numberfield.ml so functionally it's not really important. There currently is a mismatch between make_elem and get_poly though. Because int_poly does not necessarily equal A.min_poly, get_poly (make_elem p) does not necessarily equal p. I at least need some version of get_poly, and it is more convenient to just get the internal representation of a number field element. On the other hand make_elem p computes a number field element assuming p is and element of Q[x]/A.min_poly because A.min_poly is the defining polynomial the user supplied. I could (1) remove make_elem, (2) make make_elem be with respect to int_poly rather than A.min_poly, (3) make get_poly be with respect to A.min_poly rather than int_poly, or (4) leave it as is.

srk/src/numberField.ml
cyphertjohn: I don't think I understand. I tried this

let compute_min_poly_p el p = 
  let p_deg = QQX.order p in
  if p_deg = 0 then el
  else
    let e i = snd (QQX.qr (QQX.exp el i) p) in
    let rec aux app_res m i = 
      match app_res with
      | Some p -> p
      | None -> 
        let new_m = M.add_exn (e i) (QQX.exp (QQX.identity) i) m in
        aux (M.apply new_m (e p_deg)) new_m (i+1)
    in
    let min = aux None M.empty 0 in
    let lc = QQX.coeff (QQX.order min) min in
    QQX.map (fun _ c -> QQ.div c lc) min

But that doesn't seem to work. I'm thinking I'm misunderstanding M.apply and M.add

srk/src/solvablePolynomial.ml
cyphertjohn: I'm not sure I understand your point with the example. Let n = 4 for the above example. At the point in the code referenced we have:

 - inv_dom_id = <x, y>
 - transient_closure = ti^0 \cap ti^1 \cap ti^2
 - zero_eig_transient = [{x, y, z_1, z_2, z_3, z_4}, {0, 0, 0, z_1, z_2, z_3},  {0, 0, 0, 0, z_1, z_2},  {0, 0, 0, 0, 0, z_1}]
 - rels = [x', y', z_1', z_2', z_3', z_4']
 - zero_eigen_stab = 4

We then peel off elements of zero_eig_transient until we exceed iterations past the transient. The final answer is
  transient_closure \cap (<x', y', z_1', z_2', z_3', z_4' - z_1> + inv_dom_id) \cap (<x', y', z_1', z_2', z_3', z_4'> + inv_dom_id)

The middle term comes from the last element of zero_eig_transient. We only ever use the witness to approximate ti past the point of invariant domain stabilizations. For all iterations prior we take elements from the invariant domain sequence, which always is sound with respect to ti.

I've added this as a test case in test_transitionIdeal.ml. It's ultsolvable_cl2. The result is correct at least for the first 10 iterations of ti.

srk/src/solvablePolynomial.ml
cyphertjohn: Better?

srk/src/numberField.ml
zkincaid: What I had in mind was something like
```
let compute_min_poly_p el p =
    let e i = snd (QQX.qr (QQX.exp el i) p) in
    let rec aux app_res m i =
      match app_res with
     | Some p -> QQX.add_term QQ.one i p
     | None ->
       let new_m = M.add_exn (e i) (QQX.negate (QQX.exp (QQX.identity) i)) m in
         aux (M.apply new_m (e (i+1)) new_m (i+1)
    in
    aux None M.empty 0
```
The main difference being that we apply the map to `p^{i+1}` instead of `p^p_deg`.  The idea being that if p^{i+1} is in the domain of new_m. then p^{i+1} is a linear combination of p^0,...,p^i, and `M.apply new_m p^{i+1}` gives a polynomial q (of degree i) such that q(p) = -p^{i+1}.  The minimal polynomial is then x^{i+1} + q.

(I also don't understand the special-case reasoning for p_deg = 0.  I would expect the min polynomial of (say) 42 to be x-42 rather than 6, and I think the general case covers this).

srk/src/rational.ml
cyphertjohn: I've added code to handle this case. I think this should cover everything, but maybe I missed something.

srk/src/numberField.ml
cyphertjohn: I've updated compute_min_poly_p to what you have. Is there an efficiency difference between the two implementations? I'm not sure how M.add and M.apply work. Before the time was taken by a single rref call.

As to the special case for p_deg = 0, I've removed this check. It was there to handle special cases of field, but this should be handled elsewhere now.

zkincaid: I don't think a significant difference -- you can think of the map as building the RREF incrementally rather than computing the matrix and then putting it into RREF.
I'm just wary of making assumptions about the basis returned by nullspace in case its implementation changes.

srk/src/numberField.ml
zkincaid: Perhaps just rename the function (e.g., monic_min_poly_den_lcm?).  I'm still not sure that I understand, though... e.g., if A.min_poly = 4x^2 + 2, then this function gives 4, but the lcm of the denominators of (1/4)(4x^2 + 2) is 2.  (Do we assume that the denominators of A.min_poly don't have common integer factors?).

cyphertjohn: I can change the name, or we can factor the numerators of the input polynomials. It doesn't functionally matter either way. The real purpose of min_poly_den_lcm is to recognize the isomorphism between Q[x]/A.min_poly and Q[x]/int_poly. With int_poly a monic integer polynomial. I.e. defines an algebraic integer. There are many choices for int_poly that work, but we just need one. Take A.min_poly = 4x^2 + 2 then Q[x]/A.min_poly = Q(i/sqrt(2)). Taking min_poly_den_lcm = 4 will yield int_poly = x^2 + 8. So Q[x]/int_poly = Q(2isqrt(2)). Q(i/sqrt(2)) = Q(2isqrt(2)) because min_poly_den_lcm * i/sqrt(2) = 2isqrt(2). On the other hand if we want min_poly_den_lcm = 2, then int_poly = x^2 + 2. In this case Q[x]/int_poly = Q(isqrt(2)) and 2*isqrt(2) = 2isqrt(2). Everything is the same field, but just written in different bases. It doesn't matter which base we use so long as the "root" is an algebraic integer. Being an algebraic integer is only important for the order calculations later. So the point is there is no assumption on the factors of the coefficients. We could canonicalize by dividing by the lc, but as stated it doesn't matter for any calculations. In any case I think you'd just like a more descriptive name. But maybe you think we should canonicalize too. It doesn't really matter to me.

cyphertjohn: I meant denominators not numerators in the first sentence.

zkincaid: Either way is fine I think -- I just want to avoid revisiting the code in a few years and seeing that the code doesn't *exactly* match the name of the function.

srk/src/numberField.mli
zkincaid: I think it's ok as an internal function.
I'm still not quite sure that I understand what it's doing: we have two number fields
Q(a,b) and Q(c,d), represented as Q[v0,v1]/<p(v0),q(v0, v1)> and Q[v0,v1]/<r(v0),s(v0, v1)>.  We want a primitive element e such that Q(e) contains both Q(a,b) and Q(c,d).
- Is d1 the degree of Q(a,b) as a extension of Q(a) or Q?
- We only take q and s as input -- p and r are not needed?  (Do we require p = r?)
- What are v0_in_prim / v1_in_prim?

srk/src/rational.mli
```diff
@@ -0,0 +1,181 @@
...
+
+  (** Get the algebraic relations of [get_rec_sols ()]. Let d be [(Array.length (get_rec_sols ()))].
+      The result contains polynomials over the dimensions 0, ..., d-1, d, ..., 2d - 1, 2d. The
+      first 0, ..., d-1 variable are the pre-state. The d, ..., 2d-1 variables are the post-state
+      and the 2d'th variable is the iteration variable K.*)
+  val algebraic_relations : unit -> Polynomial.QQXs.t list
+
+  val shift_remove_heavys : unit -> Polynomial.QQXs.t array list * int * t array
+
+  val long_run_algebraic_relations : unit -> Polynomial.QQXs.t array list * int * Polynomial.QQXs.t list
```
zkincaid, line 141: What does this do?

srk/src/numberField.mli
cyphertjohn: Examples
 - Consider K_1 = Q(sqrt(2)) represented as Q[x]/<x^2-2> and K_2 = Q(sqrt(3)) represented as Q[y]/<y^2 - 3>. A field that contains both is K_3 = Q(sqrt(2), sqrt(3)) which can be represented by Q[x, y]/<x^2-2, y^2-3>. However, K_3 is not represented by a primitive element. The goal of the function is to find a p(z) such that K_3 = Q[z]/<p(z)>. For this example needed_deg = 4, mp0 = x^2 - 2, mp1 = y^2 - 3, v0 = x, and v1 = y. The result is p(z) = z^4 - 10z^2 + 1, which has deg, 4. v0_in_prim is a representation of x in the new field K3. For this case v0_in_prim = (z^3 - 9z)/2 and v1_in_prim = -(z^3-11z)/2. That is (v0_in_prim^2 - 2) mod <z^4 - 10z^2 + 1> = 0 and (v1_in_prim^2 - 3) mod <z^4 - 10z^2 + 1> = 0. 
 - Consider K_1 = Q[x]/<x^3-2>. Consider q(x, y) = y^2+xy+x^2 as an irreducible polynomial of K_1[y], and thus doesn't factor over K_1. We wish to enlarge K_1 with a root of q(x, y). Such a field is K3 = Q[x, y]/<x^3-2, y^2+xy+x^2>. We wish to find a primitive element that represents K3. In this case needed_deg = 6, mp0 = x^3-2, mp1 = y^2+xy+x^2, v0 = x, v1 = y. The result is p(z) = z^6 + 108. I.e. K3 = Q[z]/<z^6 + 108>. Let v0_in_prim = a(z) and v1_in_prim = b(z). Then (a(z)^3 - 2) mod <z^6 + 108> = 0 and (b(z)^2 + a(z)b(z) + a(z)^2) mod <z^6 + 108> = 0.
 - Consider K = Q[x]/<x^6 + 108>. We wish to factor y^3 - 2 in K. Consider L = K[y]/<y^3-2> = Q[x,y]<x^6 + 108, y^3-2>. L is *not* a field, because y^3 - 2 is reducible in K. Nevertheless, if we call primitive_elem 18 (x^6 + 108) (y^3 - 2) x y we will obtain a representation of L as Q[z]/p(z). As it turns out p(z) will factor over Q as three factors. Within each factor we can use v0_in_prim and v1_in_prim to get a polynomial f(x, y), which will be a factor of y^3 - 2 in K.
 
These are the three uses of primitive_elem. The first example is for "joining" fields. The second example, extends a field with an additional root. This is used in computing a splitting field. The final example doesn't compute a field at all, but is used when factoring over a number field. This is what's messy about the interface for this function. Conceptually, it is used for different things. However, all three examples perform the same computational steps.

srk/src/numberField.ml
cyphertjohn: I've changed the name

srk/src/rational.mli
cyphertjohn: I've added comments.

srk/src/numberField.mli
zkincaid: Perhaps this makes more sense in the language of commutative algebra.  Given an irreducible polynomial p in Q[v0] and a polynomial q in Q[v0,v1] \ Q[v0], we wish to find a polynomial r in Q[x] such that Q[x]/r is isomorphic to Q[v0,v1]/<p,q>.  The isomorphism is represented as v0_in_prim and v1_in_prim, which are the images of the two generators v0 + <p,q> and v_1 + <p,q> of Q[v0,v1]/<p,q> under the isomorphism.  Is that an accurate description?

cyphertjohn: Yes, that seems right.

cyphertjohn: Changed documentation.

PR #68, opened 2025-05-29
Convex hulls via local abstractions

nclskoh:
Overview of changes

- Duet is extended with options to dump (convex) hulls (-dump-hulls) for goals that call out to ConvexHull, e.g. via -cra -monotone and -termination -monotone, and an option -retype-as-real-for-lira-convhull that will compute convex hulls by first relaxing the input formula to its real relaxation.

- ConvexHull is the interface between Duet and Srk's polyhedronLatticeTiling. The implementation/interface for [abstract] is not so satisfactory because the solver's state is ignored and unchanged; the responsibility to relax the formula should perhaps move upstream to the clients.

- PolyhedronLatticeTiling implements convex hull computation for LRA, LIA, and LIRA using local abstractions and baseline algorithms. It also provides a helper function to retype/relax a formula and terms into a real formula and terms over real variables.

- Bigtop is extended with numerous options to compute convex hulls from formulas in SMT files (e.g, dumped by -dump-hulls) using different local abstraction or baseline algorithms, and to compare results between them.

Input formulas to PolyhedronLatticeTiling should only have LRA terms, but for LIRA and LIA methods, may otherwise be LIA formulas. ConvexHull and Bigtop take care of purifying formulas to ensure this.


zkincaid:
Can you submit the pull request to the flint branch to suppress the flint/modern diff?

PR #69, opened 2025-05-30
Refactor local abstraction

nclskoh:
Overview of changes

Duet is extended with options to dump (convex) hulls (-dump-hulls) for goals that call out to ConvexHull, e.g. via -cra -monotone and -termination -monotone, and an option -retype-as-real-for-lira-convhull that will compute convex hulls by first relaxing the input formula to its real relaxation.

ConvexHull is the interface between Duet and Srk's polyhedronLatticeTiling. The implementation/interface for [abstract] is not so satisfactory because the solver's state is ignored and unchanged; the responsibility to relax the formula should perhaps move upstream to the clients.

PolyhedronLatticeTiling implements convex hull computation for LRA, LIA, and LIRA using local abstractions and baseline algorithms. It also provides a helper function to retype/relax a formula and terms into a real formula and terms over real variables.

Bigtop is extended with numerous options to compute convex hulls from formulas in SMT files (e.g, dumped by -dump-hulls) using different local abstraction or baseline algorithms, and to compare results between them.

Input formulas to PolyhedronLatticeTiling should only have LRA terms, but for LIRA and LIA methods, may otherwise be LIA formulas. ConvexHull and Bigtop take care of purifying formulas to ensure this.


srk/src/dD.ml
```diff
@@ -10,6 +10,8 @@ type 'a t = 'a Abstract0.t
 type closed = Polka.loose Polka.t
 type nnc = Polka.strict Polka.t
 
+let ambient_dimension t = (Abstract0.dimension (Abstract0.manager t) t).reald
```
zkincaid: `ambient_dimension` duplicates `dimension`

srk/src/intLattice.ml
```diff
@@ -12,263 +10,328 @@ module FMPZ_mat = Flint.FMPZ_mat
...
+  type hnf
+  type unreduced
+
+  type 'form t
+
+  val bottom: unreduced t
+
+  val of_generators: V.t list -> unreduced t
+
+  val hermitize: ?compare:(int -> int -> int) -> 'form t -> hnf t
```
zkincaid: Do we really care about HNF?
In my view  is a "syntactic" property---it depends on the basis that is chosen to represent vectors, and the order that the generators are listed.  The "semantic" property that I think we do care about is whether the generators form a basis---i.e., they are linearly independent.  HNF is just one way to compute a basis.

srk/src/intLattice.ml
```diff
@@ -12,263 +10,328 @@ module FMPZ_mat = Flint.FMPZ_mat
...
-  let arr = Array.make (D.dim ctxt) ZZ.zero in
+    position of the row/matrix. *)
+
+module V = Linear.QQVector
+
+include Log.Make(struct let name = "srk.intLattice" end)
+
+let () = my_verbosity_level := `info
+
+let little_endian_list_of_vector num_dims v =
```
zkincaid: Little endian?

srk/src/intLattice.ml
```diff
@@ -12,263 +10,328 @@ module FMPZ_mat = Flint.FMPZ_mat
...
-    let ctxt = make_context dimensions in
+    (fun dim -> arr.(dim) <- V.coeff dim v)
+    (0 --^ num_dims);
+  Array.to_list arr
+
+let vector_of_list l =
+  List.fold_left (fun (v, idx) elt -> (V.add_term elt idx v, idx + 1)) (V.zero, 0) l
+  |> fst
+
+module L : sig
```
zkincaid: Just remove this module (moving its contents to the top level)?  Type signatures are in the .mli file anyway.

srk/src/syntax.mli
```diff
@@ -526,11 +535,21 @@ val lift_ite : 'a context -> 'a formula -> 'a formula
...
+    that is equivalent to [phi] when projected on to the symbols
    of [phi]. *)
 val eliminate_floor_mod_div : 'a context -> 'a formula -> 'a formula
 
+(** Given a formula [phi], compute a formula without floor, mod,
+    and div terms and without is_int atoms, whose projection onto the symbols
+    of [phi] is equivalent to [phi]. Projection must respect integrality
+    of symbols.
+ *)
+val eliminate_floor_mod_div_int : 'a context -> 'a formula -> 'a formula
```
zkincaid: What is the difference between `eliminate_floor_mod_div_int` and `eliminate_floor_mod_div`?  Why would we prefer one over the other?

srk/src/syntax.ml
```diff
@@ -1868,6 +1868,86 @@ let eliminate_floor_mod_div srk phi =
...
+  let is_ints =
+    Symbol.Set.fold
+      (fun sym l -> mk_is_int srk (mk_const srk sym) :: l
+      )
+      int_symbols
+      []
+  in
+  is_ints
+
+let retype srk (fromto: [`IntToReal | `RealToInt]) binding expr =
```
zkincaid: I  think this is simple enough to implement that we don't need a function here.
```
let retype srk table expr =
  let f symbol = match Syntax.typ_symbol with
     | `TyInt ->
           begin try Hashtbl.find table symbol
                      with Not_found ->
                              let symbol' = mk_symbol srk ~name:(Format.asprintf "%s_realified"  (show_symbol srk sym)) in
                              let k = mk_consr srk symbol' in
                              Hashtbl.add table symbol k;
                              k
           end
     | _ -> mk_const srk symbol
  in
  substitute_const srk f expr
```

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+      such that for all [pt] in [concept],
+      [f(pt)] \in [abstract pt concept] \subseteq [f(concept)].
+
+      There are three concept spaces of interest:
+      - Formulas and interpretations
+      - PLTs (polyhedron-lattice-tiling) and points ([int -> QQ.t])
+      - DDs (double-description polyhedra) and points ([int -> QQ.t])
+   *)
+
+  type ('concept1, 'point1, 'concept2, 'point2) t =
```
zkincaid: would `('point1 * 'concept1) -> ('point2 * 'concept2)` suffice? Then we could compose local abstractions using ordinary function composition.

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+    }
+
+  let conjoin lin1 lin2 =
+    {
+      p_cond = List.rev_append lin1.p_cond lin2.p_cond
+    ; l_cond = List.rev_append lin1.l_cond lin2.l_cond
+    ; t_cond = List.rev_append lin1.t_cond lin2.t_cond
+    }
+
+  let linearize_term srk vec_of_symbol term =
```
zkincaid: This generalizes `Linear.linterm_of` by adding the extra `vec_of_symbol` parameter.  I suggest adding `vec_of_symbol` this as an optional parameter to `Linear.linterm_of` that defaults to `fun k -> V.of_term QQ.one (dim_of_sym k)` and removing `linearize_term` in PLT.

srk/src/polyhedronLatticeTiling.ml
zkincaid: Or perhaps should the type be `'point -> 'concept1 -> 'concept2`?  For most local abstractions we don't really need the points to change.  For cubify, which switches from point = interpretation to point = int->QQ.t, we want the universe translation to essentially be a parameter.

srk/src/convexHull.ml
```diff
@@ -1,84 +1,58 @@
...
-        (BatDynArray.get basis i)
-    in
-    Polyhedron.local_project valuation elim_dims cube
-    |> Polyhedron.meet dim_constraint_polyhedron
-    |> Polyhedron.dd_of ~man dim
+let nb_hulls = ref 0
+let dump_hull = ref false
+let dump_hull_prefix = ref ""
+
+let lira_retype_as_real = ref false
```
zkincaid: I think it ought to be the client's responsibility realify if they wish, so that convex_hull *always* computes the convex hull.
Don't we always need formulas to be pure?

srk/src/bigtop.ml
```diff
@@ -98,6 +382,238 @@ let spec_list = [
...
    " Print generator representation of convex hull");
 
+  ("-lira-convex-hull-pc"
+  , Arg.String
+      (fun file ->
+        ConvHull.relax_to_real := NoRelax;
+        ignore
+          (ConvHull.convex_hull srk (Plt.LiraCCH PolyReccone)
+             (load_formula file));
+        Format.printf "Result: success"
```
zkincaid: Should print out the convex hull in some format

srk/src/bigtop.ml
```diff
@@ -98,6 +382,238 @@ let spec_list = [
...
+  , Arg.String
+      (fun file ->
+        ConvHull.relax_to_real := Realified;
+        ignore (ConvHull.convex_hull srk (LraCCH FullProject) (load_formula file));
+        Format.printf "Result: success"
+      )
+  , "Compute the convex hull of an existential formula in LIRA by by first expressing it as an equivalent formula in the signature of LRA using more variables, casting all variables to real, and then doing a full projection (FMCAD'15)."
+  );
+
+  ("-compare-lira-convex-hull-pc-lplh-vs-pc"
```
zkincaid: perhaps -compare-hull, which parses two text representations of convex hulls, so we don't need all the pairwise comparisons

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+
+let formula_l srk term_of_dim v =
+  let t = term_of_vector srk term_of_dim v in
+  mk_is_int srk t
+
+let formula_t srk term_of_dim v =
+  let t = term_of_vector srk term_of_dim v in
+  mk_not srk (mk_is_int srk t)
+
+let formula_of_dd srk term_of_dim dd =
```
zkincaid: Add to DD

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+         int_symbols
+         []
+
+  let cubify srk terms symbols =
+    let num_terms = Array.length terms in
+    let max_dim = get_max_dim num_terms symbols in
+    logf ~level:`debug "initial plt abstraction: max_dim: %d, num_terms = %d@;"
+      max_dim num_terms;
+    let dim_of_symbol sym = Syntax.int_of_symbol sym + num_terms in
+    let term_defs = mk_term_definitions srk dim_of_symbol terms in
```
zkincaid, line 288: It's unclear to me when we "densify" the PLT -- i.e., ensure that all of the vectors are over the dimensions [0,n] for some small n.  IntLattice does its own densification, but some polyhedra operations assume dense representation already.

I think it would be more clear if IntLattice does not do its own densification, and instead that happens inside cubify.  Cubify should be parameterized by a function that maps symbols to vectors, with the assumption that this function densifies vectors.  The `mk_term_definitions` logic can be eliminated with an appropriate choice of basis (as in https://github.com/zkincaid/duet/blob/modern/srk/src/convexHull.ml#L21)

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+    ( LocalAbstraction.{
+        abstract
+      ; translate
+      }
+    , interp_dim
+    )
+
+end
+
+module CloseStrictIneq : sig
```
zkincaid, line 338: Move module signatures + comments to .mli file

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+
+  (** Strict inequalities make projection more complex, so it helps to remove them first.
+   *)
+
+  (** For LRA (no lattice and tiling constraints), we can close strict inequalities trivially
+      because projection and taking closure commutes.
+      The lattice and tiling components of the plt are ignored and dropped.
+   *)
+  val round_assuming_no_ints: (Plt.t, int -> QQ.t, P.t, int -> QQ.t) LocalAbstraction.t
```
zkincaid: Shouldn't the input be a polyhedron?  (Assuming yes, let's move to polyhedron.ml)

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+    let bottom =
+      match bottom with
+      | None -> P.dd_of ~man target_dim P.bottom
+      | Some bot -> bot
+    in
+    let top = P.dd_of ~man target_dim P.top in
+    let srk = Abstract.Solver.get_context solver in
+    let phi = Abstract.Solver.get_formula solver in
+    let counter = ref 0 in
+    let models = ref [] in
```
zkincaid: Why save these?

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+      ; formula_of = formula_of_dd srk (fun i -> terms.(i))
+      ; top
+      ; bottom
+      }
+    in
+    Abstract.Solver.abstract solver domain
+
+end
+
+type lira_abstraction =
```
zkincaid: Why  dowe want to reify algorithm selection into a datatype?  (I.e., why not ask the client to provide the "ConvexHull.by_polyreccone" as the "how" instead of "LiraCCH PolyReccone")
Some of these configurations are for evaluation purposes only, and will eventually be removed.

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+    ?man:Polka.loose Polka.t Apron.Manager.t ->
+    'a context ->
+    'a arith_term array ->
+    Symbol.Set.t ->
+    ('a formula, 'a Interpretation.interpretation, DD.closed DD.t, int -> Q.t) LocalAbstraction.t
+
+  (** All symbols must be of integer type, and all terms must have integer coefficients.
+      Local-project-local-hull is (sound and) compact when these conditions hold.
+
+      For the latter condition, rounding assuming integer-valued variables is in general
```
zkincaid: "terms must have integer coefficients" means the terms in the `'a arith_term array`?

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1486 @@
...
+module LwCooper: sig
+
+  (** [local_project] is a simultaneous generalization of
+      Cooper-based model-based projection for linear integer arithmetic and
+      Loos-Weispfenning-based model-based projection for linear real arithmetic.
+      For linear integer-real arithmetic, [local_project] is sound (i.e., a local abstraction)
+      but may not have finite image.
+
+      Let PLT(x, Y) be a PLT in dimensions RR^{x \cup Y}.
+      [round_up: QQ^{x \cup Y} -> Term(Y) -> Term(Y)] is a function such that
```
zkincaid: Do we need round_up?  I don't understand why this is part of local projection.

srk/src/intLattice.ml
nclskoh: Legacy; was meant to convey that the smallest dimension is on the left. Renamed to list_of_vector.

srk/src/syntax.mli
nclskoh: I'm using this to make a theoretical distinction between LIA formulas and LRA formulas, and to also give LIA/LIRA local abstraction the opportunity to optimize for e.g., negative Int literals. But maybe it's more confusing this way.

srk/src/polyhedronLatticeTiling.ml
nclskoh: There's a formula_of_constraint in Polyhedron.mli, and it's easy enough to implement formula_of_dd based on that; I've dropped `formula_p` here. `formula_of_dd` here is also fully internal. I'm thinking of DD as a low-level semantic interface that currently has no dependency on Syntax; should we still move it there?

srk/src/syntax.mli
zkincaid: I think maybe I misread the comment -- I missed that is_int is also eliminated.  The resulting formula is still in LIRA, since it has integer-sorted symbols which have implicit is_int constraints.  If we absolutely need an LRA formula, we should compute its real relaxation.  Do we ever want to make explicit is_int constraints implicit?

srk/src/polyhedronLatticeTiling.ml
zkincaid: Moving the pretty-printers to DD is ok with me.  DD already has a transitive dependence on Syntax through Linear; I don't mind adding a direct dependence.

nclskoh: Added this to DD. There's a small duplication of code for formula_of_constraint to avoid mutual dependency between DD and Polyhedron, because the constraint type is the same in both.

srk/src/syntax.mli
nclskoh: This happens for computing the convex hull using baseline Gomory-Chvatal or Normaliz. We retain integer-typing for all symbols but also need to remove is_ints.

srk/src/intLattice.ml
nclskoh: Yep; do you mean to abstract this before an interface like "independent t" instead of "hnf t"?

srk/src/polyhedronLatticeTiling.ml
nclskoh: For the current implementation, `'point -> 'concept1 -> 'concept2` is indeed probably sufficient. It was packed as a data structure because universe translations were considerably more complicated with `int_frac` etc, and it was helpful to encapsulate the logic for manufacturing universe translations and composing them within factories that produce local abstractions.

Do you want me to simplify the current setup to this signature?

srk/src/polyhedronLatticeTiling.ml
nclskoh: The modules in this file including this one were meant to be internal and are used for grouping related functions. Do we want to expose them?

srk/src/polyhedronLatticeTiling.ml
nclskoh: Do we want local abstractions to be visible outside of PolyhedronLatticeTiling? Maybe another way is to have round_assuming_no_ints and round_assuming_all_ints as standard functions rather than local abstractions, with the signature Polyhedron.t -> Polyhedron.t, and have local closure in this file use that?

srk/src/polyhedronLatticeTiling.ml
nclskoh: Do you mean models? This was legacy; removed.

srk/src/polyhedronLatticeTiling.ml
nclskoh: Yeah. I will resolve some of the other comments first, then fork to have an evaluation branch that's for the experiments and submission. The one to be pulled into flint will then be trimmed down.

By "ConvexHull.by_polyreccone", do you mean to place the functions in ConvexHull? Currently ConvexHull depends on this file rather than the other way round.

srk/src/polyhedronLatticeTiling.ml
nclskoh: This is legacy for `int_frac`; I left it there because the logic for it is tricky, but I should remove it. Will update this comment when I do that.

srk/src/polyhedronLatticeTiling.ml
nclskoh: Yes. The example in the comment shows why; non-variable integrality constraints are not taken into account when we round.

srk/src/syntax.ml
nclskoh: Do you mean to replace the implementation with the one you're suggesting?

srk/src/convexHull.ml
nclskoh: What do you mean by pure? I also think that the client should be responsible for realifying, and I think we should abstract that somewhere (where?) since there are a few clients (terminationLLRF, Iteration, and bigtop). It's currently here only because I'm not sure what we want to do by default in e.g., termination, and whether that should be controlled by a flag in command-line.

srk/src/intLattice.ml
zkincaid: Yes -- I'm just thinking that it's easier for a client to understand "the generators are linearly independent" than "the matrix formed by a particular arrangement of the generators is in hermite normal form".

srk/src/syntax.ml
zkincaid: This can be specialized and inlined for the clients that need it -- e.g., computing the real relaxation of a LIRA formula is reasonable primitive that would make use of retyping, but I don't think we would ever want to replace integer symbols with reals but retain IsInt predicates.

srk/src/polyhedronLatticeTiling.ml
zkincaid: I think that makes sense.  That's the signature in the paper.

srk/src/polyhedronLatticeTiling.ml
zkincaid: Yes, we should expose local abstractions so that we can use them elsewhere.  E.g., I updated the quantifier elimination to use your PLT local abstraction, or we might want a more efficient algorithm for synthesizing quasi-ranking functions that builds on local project + hull.

srk/src/polyhedronLatticeTiling.ml
zkincaid: Can you scale up by multiplying by the LCM of the denominators and then scale back down at the end?

srk/src/polyhedronLatticeTiling.ml
zkincaid: Sounds good.
I meant that we can use `type abstraction_algorithm = Apron.Manager.t -> 'a context -> 'a term array -> symbol list` so that the `how` parameters can be the functions themselves rather than a data structure.  (The current interface is the "defunctionalization" of the interface that I'm suggesting).

srk/src/syntax.mli
zkincaid: Can we do this inside the local abstraction algorithm rather than at the formula level, so that the client needn't be responsible for it and we don't add extra symbols to the context?

nclskoh: Do you mean to eliminate floor-mod-div terms (and is_int literals if necessary) in cubify like we've done before?

srk/src/syntax.ml
nclskoh: Yep, makes sense. I will remove `retype` from `Syntax.ml`

srk/src/syntax.mli
zkincaid: Ideally, the expected pre-processing required for calling for convex hull should be the same regardless of the algorithm.  We can eliminate Gomory-Chvatal's requirement that the input is free of is_int atoms (and thus unify the algorithms pre-processing requirements for the client) by doing the transformation at the PLT level.  (And then, as a bonus, we don't have to add extra symbols to the context)

nclskoh: Good point, I can do that for GC.

duet/cra.ml
```diff
@@ -923,6 +929,9 @@ let preimage transition formula =
          mk_const srk sym
     | None -> fresh_skolem sym
   in
+  (* TODO: floor can be present in transition due to e.g., integer division.
+    Change to eliminate_floor_mod_div?
+  *)
   mk_and srk [SrkSimplify.eliminate_floor srk (K.guard transition);
```
nclskoh: Should we eliminate floor, mod and div here?

duet/cra.ml
```diff
@@ -1133,10 +1143,17 @@ let prove_termination_main file =
             let simplified =
               omega_paths_sum
               |> Nonlinear.linearize srk
-              |> Quantifier.mbp srk (fun sym ->
-                      match V.of_symbol sym with
-                      | Some x -> V.is_global x
-                      | _ -> false)
+              |> (
```
nclskoh, line 1143: `Quantifier.mbp` used to take predicate function sending symbols to keep to true and symbols to eliminate to false. This is changed to a `Symbol.Set.t`. The issue is that predicate functions can be defined to keep the complement of a subset of symbols in a formula, so when the formula is preprocessed to eliminate floor and mod by introducing some fresh symbols, they are captured by the predicate function and are not eliminated in quantifier elimination.

srk/src/abstract.ml
```diff
@@ -53,68 +51,6 @@ let boxify srk phi terms =
   | `Unsat -> mk_false srk
   | `Unknown -> assert false
 
-let abstract ?exists:(p=fun _ -> true) srk man phi =
```
nclskoh, line 56: This is moved to `srkApron.ml`.

srk/src/abstract.ml
```diff
@@ -159,19 +95,32 @@ module Solver = struct
...
 
-  let make srk ?(theory=get_theory srk) formula =
+  let to_core_lira srk ~theory phi = match theory with
+    | `LIRR -> Syntax.eliminate_floor_mod_div srk phi
+    | `LIRA ->
+      phi
+      |> Syntax.eliminate_floor_mod_div srk
+      |> Syntax.eliminate_ite srk
+      |> rewrite srk ~down:(pos_rewriter srk)
+
```
nclskoh, line 106: Preprocess a formula before it enters a solver for abstraction. The default is core LIRA, i.e., NNF with all terms in LRA and only positive inequalities. `is_int` atoms can be present (but preprocessing does not currently introduce any explicitly, only implicitly through integer-typed variables).

srk/src/abstract.ml
```diff
@@ -271,14 +220,23 @@ module Solver = struct
     | `LIRA m -> Interpretation.evaluate_formula m phi
     | `LIRR m -> Lirr.Model.evaluate_formula srk m phi
 
-  let add s phis =
+  let add s ?preprocess phis =
```
nclskoh, line 224: Expose preprocessing and do default preprocessing when adding formulas for the same reason as `make`.

srk/src/abstract.ml
```diff
@@ -655,3 +613,154 @@ let affine_hull srk phi constants =
   in
   vanishing_space srk phi basis
   |> List.map (Linear.term_of_vec srk (fun i -> basis.(i)))
+
```
nclskoh, line 617: Move `ConvexHull.ml` into `Abstract.ml` because the latter is the natural location for doing abstraction. Contract is also changed a little; see `Abstract.mli`.

srk/src/abstract.mli
```diff
@@ -19,17 +19,6 @@ val vanishing_space : 'a context -> 'a formula -> 'a arith_term array -> Linear.
     that is implied by [phi]. *)
 val boxify : 'a context -> 'a formula -> 'a arith_term list -> 'a formula
 
-(** [abstract ?exists srk man phi] computes the strongest property that is
```
nclskoh, line 22: Moved to `srkApron.mli`

srk/src/polyhedron.ml
```diff
@@ -185,90 +185,6 @@ let select_equal_term x polyhedron =
   in
   go ()
 
-(* Loos-Weispfenning virtual term *)
```
nclskoh, line 188: Superseded by PLT

srk/src/polyhedron.ml
```diff
@@ -608,116 +480,16 @@ module NormalizCone = struct
     in
     polyhedron_of (cut_eqns, cut_ineqs, bijection)
 
-  (* [fractional_cutting_planes_at_face point active_constraints ambient_dim]
```
nclskoh, line 611: Integer hull by Gomory-Chvatal moved to `DD.ml` because it requires generator representation.

srk/src/dD.ml
```diff
@@ -178,6 +192,155 @@ let minimal_faces polyhedron : (V.t * ((constraint_kind * V.t) list)) list =
       List.iter log_face defining;
       defining)
 
+module IntegerHull = struct
```
nclskoh, line 195: Moved from `Polyhedron.ml` because integer hull using Gomory-Chvatal requires generator representation.

srk/src/quantifier.ml
```diff
@@ -381,17 +234,14 @@ let is_presburger_atom srk atom =
...
+  if ZZ.equal divisor ZZ.one || V.is_zero linterm then
     mk_true srk
   else
-    let gcd = ZZ.gcd (coefficient_gcd term) divisor in
-    let divisor = QQ.of_zz (ZZ.div divisor gcd) in
-    let term = V.scalar_mul (QQ.of_zzfrac ZZ.one gcd) term in
-    mk_eq srk
-      (mk_mod srk (of_linterm srk term) (mk_real srk divisor))
-      (mk_real srk QQ.zero)
+    V.scalar_mul (QQ.inverse (QQ.of_zz divisor)) linterm
```
nclskoh, line 242: Express divisibility using `is_int` instead of mod.

srk/src/quantifier.ml
```diff
@@ -94,149 +82,14 @@ let pp_int_virtual_term srk formatter vt =
   else
     Format.fprintf formatter "@]"
 
-type virtual_term =
```
nclskoh, line 97: Logic is in `PolyhedronLatticeTiling.LwCooper`.

srk/src/srkSimplify.mli
```diff
@@ -43,26 +43,11 @@ val simplify_dda : 'a context -> 'a formula -> 'a formula
    division with a dominator less than [max]. *)
 val eliminate_idiv : ?max:int -> 'a context -> 'a formula -> 'a formula
 
-(** Purify floor functions in an expression: replace each function
```
nclskoh, line 46: Unused by clients, and there is potential for confusion with `Syntax.eliminate_floor_mod_div`.

srk/src/srkSimplify.mli
```diff
@@ -43,26 +43,11 @@ val simplify_dda : 'a context -> 'a formula -> 'a formula
...
-val purify_floor : 'a context ->
-                   ('a,'b) expr ->
-                   (('a,'b) expr * (('a,typ_arith) expr) Symbol.Map.t)
-
 (** Eliminate floor functions in a formula.  The formula is equivalent
    to the original, modulo the fresh symbols introduced in floor
    purification. *)
 val eliminate_floor : 'a context -> 'a formula -> 'a formula
 
-(** Simplify an atomic formula that consists of a binary operation of integers. *)
```
nclskoh, line 60: Using core LIRA as the language for LIRA formulas instead.

srk/src/terminationDTA.ml
```diff
@@ -250,87 +247,104 @@ let mp solver =
...
+        let cf = closed_form gz_symbols (Vec.negate v) tr_z_exp in
+        let predicate = match op with
+          | `Eq -> `Zero
+          | `Leq -> `Nonneg
+          | `Lt -> `Pos
+        in
+        XSeq.seq_of_compare_atom srk predicate cf term_of_dim
+      | `Quantify _ -> failwith "should not see quantifiers in the TF"
+      | `Atom (`ArrEq _) -> failwith "should not see ArrEq in the TF"
+      | `Atom (`IsInt t) ->
```
nclskoh, line 337: Handle divisibility here instead of in arithmetic operations above. No non-divisibility atoms are actually produced by quantifier elimination, so the case for `NotDivides` is removed.

srk/src/terminationDTA.ml
```diff
@@ -250,87 +247,104 @@ let mp solver =
...
+      | `Tru -> Periodic.make [mk_true srk]
+      | `Fls -> Periodic.make [mk_false srk]
+      | `And xs -> Periodic.mapn (mk_and srk) xs
+      | `Or xs -> Periodic.mapn (mk_or srk) xs
+      | `Not x -> Periodic.map (mk_not srk) x
+      | `Atom (`Arith (op, s, t)) ->
+        let normalize typ v =
+          if op = `Lt && (typ == `TyInt)
+          then
+            (* Strengthen inequality in integer variables with integer
```
nclskoh, line 316: Strengthening integer inequalities was part of the logic in `SrkSimplify.simplify_integer_atom`.

srk/src/transition.ml
```diff
@@ -498,8 +504,10 @@ struct
                 :: (tr_subst tr.guard)
                 :: transform_formula)
     |> Nonlinear.linearize srk
+    |> Syntax.eliminate_floor_mod_div srk
```
nclskoh, line 507: Make processing for LIRA formulas uniform.

duet/cra.ml
```diff
@@ -923,7 +929,7 @@ let preimage transition formula =
          mk_const srk sym
     | None -> fresh_skolem sym
   in
-  mk_and srk [SrkSimplify.eliminate_floor srk (K.guard transition);
+  mk_and srk [Syntax.eliminate_floor_mod_div srk (K.guard transition);
```
nclskoh, line 932: We discussed changing this to eliminating mod and div too, but I just recalled why this looked like something I wanted to change: `SrkSimplify.eliminate_floor` eliminates floor that arises in integer division for denominators up to 10 by introducing divisibility constraints, e.g., floor(x/2) --> (if (x % 2 = 0) then x / 2 else (x - 1)/2), i.e., the older version of divisibility constraints. If the divisor is > 10, a new integer-typed symbol is introduced instead.

srk/src/polyhedronLatticeTiling.ml
nclskoh: I've checked `IntLattice.ml`. Densification happens when we compute the Hermite Normal Form by calling out to Flint, and this doesn't happen when we compute convex hulls. Some time ago there was a mention about delaying HNF computation, and `IntLattice.ml` was updated then to only do HNF when needed.

srk/src/polynomialLattice.ml
```diff
@@ -5,39 +5,44 @@ module L = Log.Make(struct let name = "srk.polyLattice" end)
...
+  { ideal : Rewrite.t
+  ; affine_lattice : IntLattice.hnf IntLattice.t
+  (* Maps lowest monomial per ideal's monomial order to least integer dimension *)
   ; affine_context : LinearQQXs.context
   }
 
-let make_context monomial_order polys =
+let default_order = Monomial.degrevlex
+
+let monomials monomial_order polys =
```
nclskoh, line 18: `make_context` is factored into `monomials` and a smaller `make_context` (below).

duet/cra.ml
zkincaid: This is true, but I don't know if it's worth the overhead of needing to allocate a set of symbols to take the place of the predicate.

srk/src/abstract.ml
zkincaid: Why is the default to_core_lira?
If the general expectation is that we preprocess to get into the "core" language, I would expect rewriting to core LIRA if the background theory is LIRA and to core LIRR if the background theory is LIRR.

srk/src/abstract.ml
```diff
@@ -655,3 +613,154 @@ let affine_hull srk phi constants =
...
+          | Some vec -> BatEnum.push constraints (`Zero, vec)
+          | None -> ());
+      PolynomialCone.get_cone_generators map_cone
+      |> List.iter (fun p ->
+          match QQXs.vec_of p with
+          | Some vec -> BatEnum.push constraints (`Nonneg, vec)
+          | None -> ());
+      DD.of_constraints_closed ~man dim constraints
+
+  let print_model srk terms interp =
```
zkincaid: Expose a `pp_model` instead ( that takes a formatter as input)?

srk/src/abstract.mli
```diff
@@ -136,3 +132,46 @@ module LinearSpan : sig
...
+    formulas it can abstract, an array of terms
+    (the linear inequalities / polyhedra over which it abstracts to,
+    and a [man]ager for DD polyhedra.
+    Given [abs] that abstract to linear inequalities in (exactly) [terms],
+    a [solver] whose formula [F] can be abstracted by [abs],
+    and the [man]ager for the [abs],
+    [abstract_by man solver abs terms] is a polyhedron whose inequalities
+    when considered as inequalities in [terms] are implied by [F].
+  *)
+  val abstract_by: man:DD.closed Apron.Manager.t -> 'a Solver.t
```
zkincaid, line 165: I don't think we need to expose `abstract_by` (just `abstract`)--do you have a use case in mind?

srk/src/bigtop.ml
```diff
@@ -71,6 +70,226 @@ let print_result = function
...
+         mk_symbol srk ~name:(Format.asprintf "%s_integralized"
+                                (show_symbol srk sym))
+           `TyInt
+      end
+    in
+    let lookup s = begin try Hashtbl.find table s with
+      | Not_found ->
+        begin match (typ_symbol srk s, fromto) with
+        | (`TyInt, `IntToReal)
+          | (`TyReal, `RealToInt) ->
```
zkincaid, line 116: Can make this a little shorter with `Memo.memo retyped_symbol` (or define retyped_symbol to be `Memo.memo (fun s -> ...`).

srk/src/bigtop.ml
```diff
@@ -98,41 +317,83 @@ let spec_list = [
...
-         pp_hull (Abstract.abstract ~exists srk polka phi)),
-   " Compute the convex hull of an existential linear arithmetic formula");
+  ("-lira-convex-hull"
+  , Arg.String
+      (fun file ->
+        ignore (
+          ConvHull.convex_hull srk
+            (Plt.ConvexHull.cch_lira srk) (load_formula file)
+        );
+        Format.printf "Result: success"
```
zkincaid: I'd print the hull to stdout for each of these.

srk/src/bigtop.ml
```diff
@@ -98,41 +317,83 @@ let spec_list = [
...
+          try Retype.retype_formula srk (`LiraToLia `LraFormula) phi
+          with
+          | _ -> Format.printf "Fail at file: %s" file;
+                 failwith "Failed"
+        in
+        let suffix = if equivalent then "_equivalent.smt2" else "_integralized.smt2" in
+        let outfilename = (Filename.remove_extension file) ^ suffix in
+        Format.printf "Writing to file %s@;" outfilename;
+        let fmt = Format.formatter_of_out_channel (open_out outfilename) in
+        pp_smtlib2 srk fmt phi'
```
zkincaid: I'd print this to stdout.  Expose additional flags to determine which logical fragment a formula belongs to if need be.

srk/src/intLattice.ml
```diff
@@ -12,263 +10,270 @@ module FMPZ_mat = Flint.FMPZ_mat
...
-
-let densify ctxt vector =
-  let arr = Array.make (D.dim ctxt) ZZ.zero in
+    position of the row/matrix. *)
+
+module V = Linear.QQVector
+
+include Log.Make(struct let name = "srk.intLattice" end)
+
+let () = my_verbosity_level := `info
```
zkincaid: Delete

srk/src/iteration.ml
```diff
@@ -18,9 +18,9 @@ module Solver = struct
     ; stack : ('a Formula.t list) A.t
     ; constants : Symbol.Set.t }
 
-  let preprocess srk = function
-    | `LIRR -> Syntax.eliminate_floor_mod_div srk
-    | `LIRA -> rewrite srk ~down:(pos_rewriter srk) % (Nonlinear.linearize srk)
+  let preprocess srk theory phi = match theory with
```
zkincaid: Is the preprocessing no longer needed here because it's dones in the Abstract solver?

srk/src/nonlinear.mli
```diff
@@ -30,7 +30,13 @@ val uninterpret : 'a context -> ('a,'b) expr -> ('a,'b) expr
...
 
-(** Compute a linear approximation of a non-linear formula. *)
+(** If the theory in [ctx] is the theory of linear integer-real arithmetic (LIRA), 
+    [linearize ctx phi] is a formula in LIRA that overapproximates [phi].
+    If [phi] is already a formula in LIRA, the result is [phi] itself.
+    Formulas in LIRA have terms constructed from rational constants, addition,
+    scalar multiplication, mod constants, and floor.
+    If the theory in [ctx] is LIRR, [linearize ctx phi] is just [phi].
+*)
 val linearize : 'a context -> 'a formula -> 'a formula
```
zkincaid, line 40: We might still want to linearize if the theory is LIRR

srk/src/polyhedronLatticeTiling.mli
```diff
@@ -0,0 +1,118 @@
...
+    and integrality of symbols is taken into account by the solver.
+    The result is then a sound over-approximation of the convex hull of
+    integer-real points satisfying [F] if [is_int] only occurs positively in [F].
+
+  [F] must have only LRA terms, and hence be free of floor, mod, non-trivial
+  division, etc. This can be done by preprocessing via
+  [Syntax.eliminate_floor_mod_div], which introduces fresh integer-valued
+  variables.
+*)
+module ConvexHull : sig
```
zkincaid, line 42: I think this doesn't need to be wrapped in a module

srk/src/polyhedronLatticeTiling.mli
```diff
@@ -0,0 +1,118 @@
...
+  (** Local abstraction for core LIRA formulas, i.e., formulas whose terms
+      are LRA terms, and whose atoms are inequalities and [is_int].
+      [cch man srk symbols terms] is a local abstraction that abstracts any
+      core LIRA formula in [symbols] to a conjunction of linear inequalities over
+      [terms], i.e., a DD polyhedron, managed by [man].
+  *)
+  val cch_lira: man:DD.closed Apron.Manager.t
+    -> ?epsilon: QQ.t
+    -> 'a Syntax.context
+    -> Symbol.Set.t (* abstract formulas over this set of symbols *)
```
zkincaid: It's surprising that we would need to supply the set of symbols.  Do we need to?

srk/src/polyhedronLatticeTiling.ml
```diff
@@ -0,0 +1,1121 @@
...
+module P = Polyhedron
+module L = IntLattice
+
+module V = Linear.QQVector
+
+include Log.Make (struct let name = "srk.polyhedronLatticeTiling" end)
+
+let () = my_verbosity_level := `info
+(* let test_convex_hull = ref false *)
+let test_level = ref `debug
```
zkincaid: delete 9-11

srk/src/polyhedronLatticeTiling.ml
zkincaid: Can we still standardize on cubify returning a dense PLT?

srk/src/abstract.ml
nclskoh: Changed to `to_core`.

srk/src/abstract.mli
nclskoh: bigtop uses it to run different algorithms for LIA and LRA.

srk/src/iteration.ml
```diff
@@ -18,14 +18,9 @@ module Solver = struct
...
 
-  let preprocess srk = function
-    | `LIRR -> Syntax.eliminate_floor_mod_div srk
-    | `LIRA -> rewrite srk ~down:(pos_rewriter srk) % (Nonlinear.linearize srk)
-
   let make srk ?(theory=get_theory srk) tf =
-    let phi = preprocess srk theory (TF.formula tf) in
     let stack = A.singleton [TF.formula tf] in
-    { solver = Abstract.Solver.make srk ~theory phi
+    { solver = Abstract.Solver.make srk ~theory (TF.formula tf)
```
nclskoh, line 23: Removed preprocessing from `make` and `add` (below).

srk/src/nonlinear.mli
nclskoh: Yup, but the current implementation is identity on LIRR, and I was thinking the comment should give some specification for LIRR since we mention one for LIRA.

srk/src/polyhedronLatticeTiling.mli
nclskoh: Removed symbols; they were used earlier because new dimensions were introduced to remove floor and mod.

srk/src/abstract.ml
nclskoh: I've removed it. I think `Interpretation` and `Lirr` expose enough to print what's needed for each theory based on client's needs, and the function is simple enough to implement.

srk/src/bigtop.ml
nclskoh: The hash table is returned for the caller to remap existentially quantified symbols to the new ones.

srk/src/polyhedronLatticeTiling.mli
nclskoh: I'm thinking of having another module for affine integer span. Do you still prefer removing the module? Also, what do you think about the comments?

srk/src/polyhedronLatticeTiling.ml
nclskoh: By standardization, do you mean that over [here](https://github.com/nclskoh/duet/blob/c7aed1a4d7b162c4e564022312b2082c5b0a094d/srk/src/intLattice.ml#L133-L134), we should not remap and just assume that vectors are already dense with all dimensions in the right places?

There's a remap that's required for [projection](
https://github.com/nclskoh/duet/blob/c7aed1a4d7b162c4e564022312b2082c5b0a094d/srk/src/intLattice.ml#L215-L221), but if the concern is with performance, I can match on whether a conversion is supplied to do a remap, and otherwise assume all dimensions are in the right places.

srk/src/bigtop.ml
nclskoh: Good point. What additional flags do you mean?

PR #71, opened 2025-07-15
GPS 

ruijiefang:


duet/summaryProvider.ml
```diff
@@ -0,0 +1,244 @@
+open Core
```
zkincaid, line 1: Dead code?

duet/summaryProvider.mli
```diff
@@ -0,0 +1,159 @@
+(*
```
zkincaid, line 1: Dead code?

srk/src/transition.ml
```diff
@@ -113,40 +115,47 @@ struct
...
-    let guard =
-      mk_or srk [mk_and srk (left.guard::(!left_eq));
-                 mk_and srk (right.guard::(!right_eq))]
-    in
-    { guard; transform }
+      let guard = match ty with
+      | `Add ->
+        mk_or srk [mk_and srk (left.guard::(!left_eq));
+                    mk_and srk (right.guard::(!right_eq))]
+      | `And ->
```
zkincaid, line 150: Skolem constants must be renamed for And (e.g., "K.assume(sk1 = 0) and K.assume(sk1 = 1)" should the same as "K.assume(sk1 = 0 /\ sk2 = 1)", rather than false).
And also doesn't require introducing phi variables -- we can simply equate the right-hand sides of the transform (e.g., [x -> 0] /\ [x -> y] --> [x -> 0 when y = 0]).

srk/src/transition.ml
```diff
@@ -438,10 +447,459 @@ struct
              (wp::itp, wp))
            trs
            guards
-           ([post], post)
+          ([
```
zkincaid, line 450: Maybe cleaner to require post to be a state formula that is free of Skolem constants?  Then rather using the guard of the path-to-error summary in GPS as the post-condition, we place the path-to-error as the last transition in trs and the post-condition is false.  (Maybe there's some performance overhead with this encoding?)

srk/src/transition.ml
```diff
@@ -438,10 +447,459 @@ struct
...
+          ) (Interpretation.wrap srk (fun s ->
+              match Hashtbl.find_opt ss s with
+              | Some sss -> Interpretation.value model sss
+              | None -> `Real (Q.of_int 47))) (*(Interpretation.wrap srk (fun s ->
+                match Hashtbl.find_opt ss s with
+                | Some sss -> Interpretation.value model sss
+                | None -> Interpretation.value model s))*) (*(Interpretation.empty srk)*) symbols in
+          logf "hashtable length: %d\n" (Hashtbl.length ss_inv);
+          logf "%a" Interpretation.pp m;
+          Format.print_flush ();
```
zkincaid, line 663: logf flushes

srk/src/transition.ml
zkincaid: Whatever overhead there is might be worth it, because interpolate_or_get_model doesn't generate indicator variables for the post-condition, just the transitions.  So the post-condition is *always* the guard of the path-to-error summary, I think.  We generally want to allow it to be stronger.

srk/src/transition.ml
```diff
@@ -438,10 +447,459 @@ struct
...
+          ([], ss_guards)
+      in
+      List.iter (fun (k, l, v) ->
+        Hashtbl.add subscript_tbl k v;
+        Hashtbl.add ss_inv l k;
+        Hashtbl.add sst k l) ss;
+      mk_and srk phis
+    in
+    (* gather all symbols into a list, while adding formulas to the solver object *)
+    let symbols, added_formulas = List.fold_left
```
zkincaid, line 599: Do we need "added_formulas"?

srk/src/transition.ml
```diff
@@ -438,10 +447,459 @@ struct
...
+          Format.print_flush ();
+      (* symbols is a list of subscripted symbols arranged in left-to-right order.
+         folding over this in left-to-right order amounts to forward concrete execution. *)
+      `Invalid (m
+        |> Interpretation.restrict
+          (fun s ->
+            match Var.of_symbol s with
+            | Some _ -> true
+            | None -> false))
+    in interpolate_query trs post sat_model @@ interpolate_unsat_core
```
zkincaid, line 672: Is this the only client of interpolate_query?  If yes, can we simplify the logic?

srk/src/transition.ml
```diff
@@ -438,10 +447,459 @@ struct
...
+  let vocabulary tr =
+    let tr_guard = guard tr in
+    let tr_trans = transform tr in
+    let guard_v = tr_guard |> Syntax.symbols in
+    let trans_v = BatEnum.fold (fun s (var, term) ->
+      let s = Symbol.Set.add (Var.symbol_of var) s in
+      let t = Syntax.symbols term in
+      Symbol.Set.union s t) Symbol.Set.empty tr_trans in
+    let v = Symbol.Set.union guard_v trans_v in
+    let globals = Symbol.Set.filter (fun x ->
```
zkincaid, line 684: Symbol.Set.partition gets both.  Or use to_list first, then List.partition

srk/src/transitionSystem.ml
```diff
@@ -662,18 +684,168 @@ module Make
...
+          begin match PHT.find_opt qmap (x, y) with 
+          | None -> 
+            PHT.add qmap (x, y) PPS.empty 
+          | Some _ -> () 
+          end;
+          PS.add (x, y) acc
+        | _ -> acc 
+        ) tg (PS.add (entry, -1) PS.empty) in 
+    let rec dfs (proc: int * int) tg src (f: (vertex * vertex) -> vertex -> vertex * 'a label * vertex -> unit) (visited : ISet.t) = 
+      Printf.printf "dfs: visiting %d\n" src;
```
zkincaid, line 713: Use ``logf`` so that we suppress this output by default

srk/src/transitionSystem.ml
```diff
@@ -662,18 +684,168 @@ module Make
     in
     List.map invariants (L.all_loops (L.loop_nest tg))
 
-  let simplify p tg =
+
+  let inline ?(depth=(-1)) tg entry (displayer: t -> unit) (assertion_map : 'x SrkUtil.Int.Map.t) = 
+    let pmap = PHT.create 998 in (* y in pmap[x] means call from x->y *)
```
zkincaid, line 689: ``WeightedGraph.RecGraph`` has a ``CallGraph`` data structure that you can use to create callgraphs.  OCamlgraph has ``Components`` and ``Topological`` modules that you can use to find SCCs/traverse the call graph in topological order.

srk/src/transitionSystem.ml
```diff
@@ -662,18 +684,168 @@ module Make
...
+      let copy_subgraph tg src = 
+        let vertices = dfs (-1, -1) tg src (fun _ _ _ -> ()) (ISet.add src ISet.empty) in 
+        let to_map = Hashtbl.create 998 in 
+        let rtg = ref tg in 
+          ISet.iter (fun x -> 
+            greatest := !greatest + 1;
+            Hashtbl.add to_map x (!greatest);
+            begin match SrkUtil.Int.Map.find_opt x !assertions with (* if x in assertions, new vertex also in assertions *)
+              | Some assert_value -> 
+                assertions := SrkUtil.Int.Map.add !greatest assert_value !assertions
```
zkincaid, line 752: In my view, there are two possible types of clients
1. We have a set of assertions in the program, and we want to prove or disprove each separately.
2. We don't care to distinguish between asserts -- we want to prove all of them or refute one of them.
In case 1, we shouldn't duplicate asserts because, for instance, the client won't be able to make sense of an assert being both safe and unsafe (which can happen if the function containing the assert is duplicated for two different call sites).  In case 2, we also shouldn't duplicate asserts because we want to maintain the invariant that there's only one.

I suggest that we add a comment to inline that we must have the pre-condition that assert nodes have no outgoing edges, and inline should only copy non-assert nodes.

srk/src/transitionSystem.ml
```diff
@@ -662,18 +684,168 @@ module Make
     in
     List.map invariants (L.all_loops (L.loop_nest tg))
 
-  let simplify p tg =
+
+  let inline ?(depth=(-1)) tg entry (displayer: t -> unit) (assertion_map : 'x SrkUtil.Int.Map.t) =
```
zkincaid, line 688: Is displayer just for debugging inline?  If yes, we can omit this parameter now.

srk/src/transitionSystem.ml
```diff
@@ -662,18 +684,168 @@ module Make
...
+          | Some callees, Some callers -> 
+            (* a procedure is considered for inlining if it is (1) a sink in the call graph (2) at least one function calls it.*)
+            if ((PS.cardinal callees) == 0) && ((PPS.cardinal callers) > 0) then PS.add (x, y) acc else 
+              begin 
+                Printf.printf "%d %d is not a sink; num callees = %d num callers = %d\n" x y (PS.cardinal callees) (PPS.cardinal callers);
+                acc
+              end 
+          | (_, _) -> failwith ""
+          ) procedures PS.empty in
+      let copy_subgraph tg src =
```
zkincaid, line 743: copy_subgraph doesn't rename variables, so it's unsafe for recursive procedures.  That's ok--complete inlining won't work anyway---but add a comment to the inline function that recursion is not supported, even with the depth bound.

duet/cra.ml
```diff
@@ -841,7 +976,12 @@ let make_transition_system rg =
       TS.empty
       (RG.bodies rg)
   in
-  (ts, !assertions)
+  (* perform some inlining *)
+  Printf.printf "calling inliner...\n";
```
zkincaid, line 980: Use ``logf`` rather than ``printf``

duet/cra.ml
```diff
@@ -751,21 +787,114 @@ let decorate_transition_system predicates ts entry =
...
+  let gas_var_sym = Syntax.mk_symbol srk ~name:"__duet_gas" `TyInt in  
+  let gas_var_term = Syntax.mk_const srk gas_var_sym in 
+  Hashtbl.add V.sym_to_var gas_var_sym (VVal gas_var);
+  ValueHT.add V.var_to_sym (VVal gas_var) gas_var_sym;
+  let gasexpr = (Syntax.mk_lt srk (mk_int 0) gas_var_term) in 
+  let gasweight =
+    let assume_positive = K.assume gasexpr in
+    let decr_by_one = Syntax.mk_sub srk gas_var_term (mk_int 1) |> K.assign (VVal gas_var) in
+      K.mul assume_positive decr_by_one in 
+  let initial_gas = K.havoc [ VVal gas_var ] in
```
zkincaid, line 811: Do we need the havoc?  Should have the same effect if we leave it uninitialized.

duet/cra.ml
```diff
@@ -751,21 +787,114 @@ let decorate_transition_system predicates ts entry =
...
+    let assume_positive = K.assume gasexpr in
+    let decr_by_one = Syntax.mk_sub srk gas_var_term (mk_int 1) |> K.assign (VVal gas_var) in
+      K.mul assume_positive decr_by_one in 
+  let initial_gas = K.havoc [ VVal gas_var ] in  
+  (gasweight, gasexpr, initial_gas) 
+
+let new_vtx () = (Def.mk (Assume Bexpr.ktrue)).did 
+
+let instrument_with_gas (ts: TSG.t) gasexpr : TSG.t = 
+  let modify_pre ts u =
```
zkincaid, line 817: There's a convenient function in ``WeightedGraph`` that helps with this kind of instrumentation---``split_vertex wg u weight v`` takes all of the outgoing edges of u and makes them outgoing edges of v, and adds an edge from u to v with the given weight.

duet/gps.ml
```diff
@@ -0,0 +1,747 @@
...
+  List.iteri (fun i f -> logf "[formula] %s(%i): %a\n" prefix i (Syntax.pp_expr srk) f) formulas
+
+let log_weights prefix weights =
+  List.iteri (fun i f -> logf "[weight] %s(%i): %a\n" prefix i K.pp f) weights
+
+
+let log_model prefix model =
+  logf "[model] %s: %a\n" prefix Interpretation.pp model
+
+(*
```
zkincaid, line 54: Dead code?

duet/gps.ml
```diff
@@ -0,0 +1,747 @@
...
+    SrkUtil.Int.Map.fold (fun v (phi, _, _) ts ->
+        let s = Printf.sprintf " Adding assertion node %d -> %d for label " v err_loc in
+        log_formulas s [ Ctx.mk_not phi ] ;
+        WG.add_edge ts v (Weight (K.assume (Ctx.mk_not phi))) err_loc)
+      assertions
+      ts
+  in
+  (ts, err_loc)
+
+module Summarizer =
```
zkincaid, line 84: I think we can omit all of the interprocedural logic until we have a better idea of what it's going to look like in the end.

duet/reachTree.ml
```diff
@@ -0,0 +1,616 @@
...
+module Interpretation = Srk.Interpretation 
+
+
+include Log.Make (struct
+  let name = "reachTree"
+end)
+
+type equery = OverApprox | UnderApprox
+
+module ART
```
zkincaid, line 22: Call this functor ``Make``?

ruijiefang:
Oops --- sorry Zak, I should have opened a PR from this branch: https://github.com/ruijiefang/duet/tree/gps-improvements, or at least merge changes there into the `gps` branch.

ruijiefang:
Let us continue the discussion at https://github.com/zkincaid/duet/pull/72

PR #73, opened 2025-09-17
Extended transition.ml to support array terms

nikhilpim:


duet/cra.ml
```diff
@@ -485,13 +489,13 @@ let rec record_assign (lhs : varinfo) loff rhs roff fields =
         begin match tr_expr (AccessPath (Variable (rhs, OffsetFixed roff))) with
           | TPointer rhs ->
             BatList.reduce K.mul [
-              K.assign (VVal lhs) rhs.ptr_val;
-              K.assign (VPos lhs) rhs.ptr_pos;
-              K.assign (VWidth lhs) rhs.ptr_width;
+              K.assign (VVal lhs) (Syntax.Term.promote_arith srk rhs.ptr_val);
```
zkincaid: Given the number of calls of this form, it would clean things up by adding a "arith_assign" function to transition.ml with the appropriate type.

Also, `promote_arith` is not needed to convert the phantom type since arithmetic terms are a sub-type of terms.  We do need an explicit up-cast, however:
`(Syntax.Term.promote_arith srk rhs.ptr_val)`
is the same as
`(rhs.ptr_val :> 'a Syntax.term)`.

srk/src/transition.ml
```diff
@@ -136,8 +135,12 @@ struct
             | Some t -> t
             | None -> mk_const srk (Var.symbol_of v)
           in
-          left_eq := (mk_eq srk left_term phi)::(!left_eq);
-          right_eq := (mk_eq srk right_term phi)::(!right_eq);
+          left_eq := (match Term.refine srk left_term with
```
zkincaid: Here and a couple of other places we have the same logic, in which we *want* to equate two terms regardless of type but can't.  I suggest extracting this logic to a single function in syntax.ml.

PR #74, opened 2026-02-02
Map Elimination Implementation

nikhilpim:
array_lift.ml - implementation of the "bubble" algorithm for map elimination, and two handlers: map_elim which eliminates array variables from a formula and array_exponentiate which lifts exponentiation operators to work over arrays.
test_arraylist.ml - some initial tests for these functions
transitionFormula - added a printing function for transition formulas.


srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+
+(* bubble_sym is an internal helper function for lowering formulas in the skolem array fragment
+    to quantified numerical formulas. 
+    [bubble_sym srk form] returns a triple of the form (syms, name, form') in which:
+        - form ==> \exists syms. \forall name. form'
+        - form' contains no array terms
+*) 
+let bubble_sym (srk : 'a context) (form : 'a formula) : (symbol list * (string option) * 'a formula) = 
+  (* rep_map : arr_var -> (index_var -> var) *)
+  let (rep_map : (symbol SMap.t) SMap.t ref) = ref SMap.empty in
```
zkincaid: Could use a hash table.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+
+  (* All variables are turned into symbols as we recurse across the formula.
+    env maps from variable indices to symbols. 
+    qo is the index of the forall variable. 
+    Variables with index > qo are existentially quantified outside the forall.
+    Note that replace functions only get run after we have found the forall.
+  *)
+  (* [replace_access arr i env qo] returns a substitute term for arr[i] *)
+  let rec replace_access (arr : 'a arr_term) (index : 'a arith_term) (env : (symbol Env.t)) (qo : int) : 'a arith_term = 
+    print_string ("Replacing array access: " ^ (ArrTerm.show srk arr) ^ "[" ^ (ArithTerm.show srk index) ^ "]\n");
```
zkincaid: Don't use print_string for debug info -- omit or use `logf ~\level:trace` (and start the file with `include Log.Make(struct let name = "Srk.ArrayLift" end` so that logf is defined).

srk/src/transitionFormula.ml
```diff
@@ -7,6 +7,18 @@ type 'a t =
 
 include Log.Make(struct let name = "srk.transitionFormula" end)
 
+let show srk tf =
```
zkincaid: Export a pretty printer rather than convert to string (see, e.g., the pp function in transition.ml)

srk/src/arraylift.mli
```diff
@@ -0,0 +1,9 @@
+open Syntax
+
+
+val array_exponentiate :
+  'a context ->
+  'a Iteration.exp_op ->
+  ('a TransitionFormula.t -> 'a arith_term -> 'a formula)
+
+val map_elim : 'a context -> 'a formula -> 'a formula
```
zkincaid, line 17: Add comments following ocamldoc style

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+    | `Atom (`ArrEq _) -> failwith "Array equalities should not appear in Skolem fragment"
+    | `Not _ -> failwith "Not is not supported in Skolem fragment"
+  in
+
+  (* go contains the core "bubbling" logic of this algorithm: 
+    we process leading existentials, skolemizing them and adding them to env.
+    Once we reach a forall, we stop processing existentials and switch over to replace functions
+    to eliminate array terms.
+  *)
+  let rec go (f : 'a formula) (env : (symbol Env.t)): (symbol list * string option * 'a formula) =
```
zkincaid: Document what `go` returns

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+        This has a consequence at line 30. 
+      *)
+      ([], None, replace f env 0)
+    | `Tru | `Fls -> ([], None, f)
+    | `Ite (f, l, r) -> 
+      let f' = mk_or srk [mk_and srk [f ; l]; mk_and srk [mk_not srk f; r]] in 
+      go f' env
+    | `Not _ -> failwith "Not is not supported in Skolem fragment"
+    in 
+    print_string "\nBubbled to :\n"; print_string (Formula.show srk (List.fold_left (fun acc sym -> mk_exists_const srk sym acc) (match forall with | None -> retf | Some name -> mk_forall srk ~name `TyInt retf) syms)); print_newline ();
```
zkincaid: I suspect this prints too much -- I'd remove from the PR and re-introduce locally when needed for debugging.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+      let array_symbol = Env.find env (i - qo) in 
+      let replacement_map = 
+        try SMap.find array_symbol !rep_map 
+        with Not_found -> failwith "Array variable was never existentially quantified"
+      in 
+      match ArithTerm.destruct srk index with 
+        | `Var (indi, _) -> 
+          ((if indi < qo then failwith "Array index variable is bound but should be free");
+          let index_symbol = Env.find env (indi - qo) in 
+          if (SMap.mem index_symbol replacement_map) then (
```
zkincaid: Factor this logic out into a helper function, and call it here and line 48.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+    to eliminate array terms.
+  *)
+  let rec go (f : 'a formula) (env : (symbol Env.t)): (symbol list * string option * 'a formula) = 
+    let  syms, forall, retf = match Formula.destruct srk f with 
+    | `Quantify (`Exists, name, typ, body) -> 
+      let new_sym = mk_symbol srk ~name (typ :> typ) in 
+      (if typ = `TyArr then (rep_map := (SMap.add new_sym (SMap.empty) !rep_map))); 
+      let (syms, forall, body) = go body (Env.push new_sym env) in
+      if typ = `TyArr then (
+        let introduced_constants = SMap.find new_sym !rep_map |>
```
zkincaid: If the range of rep_map always contains Skolem constants, you can simplify this logic by maintaining a mutable list of all skolem constants and adding a constants to it in your [helper function](https://github.com/zkincaid/duet/pull/74/changes#r2763912503)

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+      if typ = `TyArr then (
+        let introduced_constants = SMap.find new_sym !rep_map |> 
+          SMap.bindings |> List.map snd in 
+        (introduced_constants @ syms, forall, body)
+      ) else (new_sym :: syms, forall, body) 
+      
+    | `Quantify (`Forall, name, typ, body) -> 
+      assert (typ = `TyInt);
+      let (body) = replace body (Env.push forall_symbol env) 0  in 
+      ([], Some name, body)
```
zkincaid: Names are not necessarily consistent across all foralls, so I wouldn't try to compute one.  I think we just need a boolean to indicate whether we've seen a universal quantifier.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+      (* At this point, we want to run replace but cannot because we haven't yet seen a forall.
+        Ideally, we could use the equivalence f <=> \forall d. f where f is d-free.
+        However, this involves renaming free variables within f to avoid conflict. 
+
+        Instead, we can use a hacky workaround: just running replace as normal. 
+        This has a consequence at line 30. 
+      *)
+      ([], None, replace f env 0)
+    | `Tru | `Fls -> ([], None, f)
+    | `Ite (f, l, r) ->
```
zkincaid, line 124: This is going to crash because Not isn't supported -- might as well crash here rather than transform.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+      mk_is_int srk t'
+    | `Atom (`Arith (op, t1, t2)) -> 
+      (let (t1') = replace_at t1 env qo in 
+      let (t2') = replace_at t2 env qo in 
+      match op with 
+        | `Eq -> (mk_eq srk t1' t2')
+        | `Leq -> (mk_leq srk t1' t2')
+        | `Lt  -> (mk_lt srk t1' t2'))
+    | `Tru | `Fls | `Proposition _ -> f
+    | `And ls -> mk_and srk (List.map (fun f -> replace f env qo) ls)
```
zkincaid: make f the last parameter to replace, so that this is just `(List.map (replace env qo) ls)`

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+    let  syms, forall, retf = match Formula.destruct srk f with 
+    | `Quantify (`Exists, name, typ, body) -> 
+      let new_sym = mk_symbol srk ~name (typ :> typ) in 
+      (if typ = `TyArr then (rep_map := (SMap.add new_sym (SMap.empty) !rep_map))); 
+      let (syms, forall, body) = go body (Env.push new_sym env) in
+      if typ = `TyArr then (
+        let introduced_constants = SMap.find new_sym !rep_map |> 
+          SMap.bindings |> List.map snd in 
+        (introduced_constants @ syms, forall, body)
+      ) else (new_sym :: syms, forall, body)
```
zkincaid: This seems to be missing
* existential quantifier for the variable introduced to represent new_sym[i] (where i is the universally quantified variable).  This can't be skolemized, because the existential must appear *inside* the universal
* Logic for functional consistency (e.g., exists a. x = y /\ select(a,x) != select(a,y) is unsat, but if you replace select(a,x) and select(a,y) with skolem constants, the result is sat).

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+        | _ -> failwith "Array index should be a variable or a symbol"
+      ) 
+        else failwith "Array term is not quantified inside the forall"
+    | `App (_, _) -> failwith "Array applications not supported"
+    | `Ite (f, l, r) -> mk_ite srk (replace f env qo) (replace_access l index env qo) (replace_access r index env qo)
+    | `Store (a, store_index, value) ->
+      let ae = replace_access a index env qo in 
+      mk_ite srk (mk_eq srk store_index index) value ae
+  (* replace_at recurses over arith_terms, performing substitution *)
+  and replace_at (t : 'a arith_term) (env : (symbol Env.t)) (qo : int) : 'a arith_term =
```
zkincaid: I think this is: `rewrite srk ~up:(function Select(arr, index) -> replace_access arr index env qo | x -> x) t`

srk/test/test_arraylift.ml
```diff
@@ -0,0 +1,117 @@
...
+    (fun acc s -> mk_exists_const srk s acc) combined [asym; asym'; bsym; bsym'; dsym; dsym'; ksym; ksym'; nsym; nsym'] in 
+  let pos = rewrite srk ~down:(pos_rewriter srk) quantified in 
+  let equi_sat = Arraylift.map_elim srk pos in 
+  match Quantifier.simsat srk equi_sat with
+  | `Sat -> print_string "Sat\n"
+  | `Unsat -> print_string "Unsat\n"
+  | `Unknown -> print_string "Unknown\n"
+
+
+let suite = "Iteration" >::: [
```
zkincaid, line 219: Should have test cases exercising
* array accesses where the accessed index is an existentially quantified variable or constant symbol
* array accesses that do not appear inside the scope of a universal
* array store

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+          ))
+        | `App (s, []) -> 
+          if (SMap.mem s replacement_map) then (
+            mk_const srk (SMap.find s replacement_map))
+           else (
+            let replacement = mk_symbol srk ~name:("rep_" ^ (show_symbol srk array_symbol) ^ "_" ^ (show_symbol srk s)) `TyInt in
+            rep_map := SMap.add array_symbol (SMap.add s replacement replacement_map) !rep_map;
+            mk_const srk replacement
+           )
+        | _ -> failwith "Array index should be a variable or a symbol"
```
zkincaid: I'd swap the "then" and "else" branches of this if so that the short case comes first -- easier to read.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,197 @@
...
+      ) 
+        else failwith "Array term is not quantified inside the forall"
+    | `App (_, _) -> failwith "Array applications not supported"
+    | `Ite (f, l, r) -> mk_ite srk (replace f env qo) (replace_access l index env qo) (replace_access r index env qo)
+    | `Store (a, store_index, value) ->
+      let ae = replace_access a index env qo in 
+      mk_ite srk (mk_eq srk store_index index) value ae
```
zkincaid: Don't we need to perform a substitution on index?

srk/src/arraylift.ml
nikhilpim: Changed to apply pos_rewriter (which I think will allow the algorithm to continue).

srk/src/arraylift.ml
nikhilpim: There's some details in the handling of Var and Ite that aren't so easy to express as a rewriter - I think it's more readable to leave everything explicit.

srk/test/test_arraylift.ml
nikhilpim: added new tests.

nikhilpim:
@zkincaid I'm done making edits based on your review - it's ready to be looked over again. All tests pass.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+  (* rep_map : arr_var -> (index_term -> var) *)
+  let (rep_map : (symbol, ('a arith_term, symbol) Hashtbl.t) Hashtbl.t) = Hashtbl.create 16 in 
+  let created_symbols = ref [] in 
+  let forall_symbol = mk_symbol srk `TyInt ~name:"forall_index" in
+
+  let get_or_create_replacement srk replacement_map array_symbol index_term isnt_forall = 
+    match Hashtbl.find_opt replacement_map index_term with 
+      | Some replacement -> mk_const srk replacement
+      | None -> (
+        let replacement = mk_symbol srk `TyInt ~name:("rep_" ^ (show_symbol srk array_symbol) ^ "_" ^ (ArithTerm.show srk index_term)) in
```
zkincaid: Format.asprintf "rep_%a_%a" Symbol.pp (pp_symbol srk) array_symbol (ArithTerm.pp srk) index_term

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+(* bubble_sym is an internal helper function for lowering formulas in the skolem array fragment
+    to quantified numerical formulas. 
+    [bubble_sym srk form] returns a triple of the form (syms, is_forall, form') in which:
+        - form ==> \exists syms. \forall "". form' if is_forall 
+        - form ==> \exists syms. form' if not is_forall
+        - form' contains no array terms
+*) 
+let bubble_sym (srk : 'a context) (form : 'a formula) : (symbol list * (symbol option) * 'a formula) = 
+  (* rep_map : arr_var -> (index_term -> var) *)
+  let (rep_map : (symbol, ('a arith_term, symbol) Hashtbl.t) Hashtbl.t) = Hashtbl.create 16 in
```
zkincaid: arith_term should not be used as a key for hashtbl -- use Expr.HT (this gives constant-time hashing).
(Or use Expr.ExprMemo to memoize get_or_create_replacement)

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+        - form' contains no array terms
+*) 
+let bubble_sym (srk : 'a context) (form : 'a formula) : (symbol list * (symbol option) * 'a formula) = 
+  (* rep_map : arr_var -> (index_term -> var) *)
+  let (rep_map : (symbol, ('a arith_term, symbol) Hashtbl.t) Hashtbl.t) = Hashtbl.create 16 in 
+  let created_symbols = ref [] in 
+  let forall_symbol = mk_symbol srk `TyInt ~name:"forall_index" in
+
+  let get_or_create_replacement srk replacement_map array_symbol index_term isnt_forall = 
+    match Hashtbl.find_opt replacement_map index_term with
```
zkincaid: Why is replacement_map a parameter?

srk/src/arraylift.ml
zkincaid: What are the issues?

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+     match Formula.destruct srk f with 
+    | `Quantify (`Exists, name, typ, body) -> 
+      let new_sym = mk_symbol srk ~name (typ :> typ) in 
+      (if typ = `TyArr then ((Hashtbl.add rep_map new_sym (Hashtbl.create 16)))); 
+      let (syms, forall, body) = go body (Env.push new_sym env) in
+      if typ = `TyArr then (syms , forall, body) else (new_sym :: syms, forall, body) 
+    | `Quantify (`Forall, _, typ, body) -> 
+      assert (typ = `TyInt);
+      let (body) = replace (Env.push forall_symbol env) 0 body in 
+      let functional_correctness = (Hashtbl.fold (fun array_symbol m (acc) ->
```
zkincaid: functional *consistency*

zkincaid: The logic is not correct here.  Functional consistency axioms for the array a need to be inserted at the existential for a, not the universal quantifier.

Test case that should reveal the problem: exists n,m : int. exists a : array. n = m /\  a[n] != a[m]

srk/src/arraylift.ml
zkincaid: srk should also not be a parameter

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+    | `Quantify (`Exists, name, typ, body) -> 
+      let new_sym = mk_symbol srk ~name (typ :> typ) in 
+      (if typ = `TyArr then ((Hashtbl.add rep_map new_sym (Hashtbl.create 16)))); 
+      let (syms, forall, body) = go body (Env.push new_sym env) in
+      if typ = `TyArr then (syms , forall, body) else (new_sym :: syms, forall, body) 
+    | `Quantify (`Forall, _, typ, body) -> 
+      assert (typ = `TyInt);
+      let (body) = replace (Env.push forall_symbol env) 0 body in 
+      let functional_correctness = (Hashtbl.fold (fun array_symbol m (acc) -> 
+          let ai = get_or_create_replacement srk m array_symbol (mk_const srk forall_symbol) false in
```
zkincaid: Since you're always creating this ai symbol, might as well insert it into replacement_map when we see the existential for array_symbol.  Then you can remove the bool parameter to get_or_create_replacement

srk/src/arraylift.ml
```diff
@@ -0,0 +1,211 @@
...
+    we process leading existentials, skolemizing them and adding them to env.
+    Once we reach a forall, we stop processing existentials and switch over to replace functions
+    to eliminate array terms.
+  *)
+  (* [go form env] considers the formula \exists env . form
+     and returns (syms, is_forall, form') where
+        - \exists env . form ==> \exists syms. \forall name. form' if is_forall
+        - \exists env . form ==> \exists syms. form' if not is_forall
+        - form' contains no array terms *)
+  let rec go (f : 'a formula) (env : (symbol Env.t)): (symbol list * bool * 'a formula) =
```
zkincaid: We don't need the symbol_list return -- isn't that what created_symbols is for?

srk/src/arraylift.mli
```diff
@@ -0,0 +1,17 @@
...
+  , and returns an exponentiation operator which numerically abstracts the array variables
+  in the formula, applies exp_op, then reintroduces the array variables 
+*)
+val array_exponentiate :
+  'a context ->
+  'a Iteration.exp_op ->
+  ('a TransitionFormula.t -> 'a arith_term -> 'a formula)
+
+
+(** [map_elim ctx formula] computes an equisatisfiable formula to [formula]
```
zkincaid, line 14: Equivalent, not just equisatisfiable.

srk/src/arraylift.ml
nikhilpim: Well, Var (i, t) needs to search in the environment to find the relevant skolem constant, which is easy to put in a rewriter.

But Ite (f, l, r) needs to apply replace (for formulas) to f and replace_at to l and r. I'm not sure exactly how to put that as a rewriter.

Does changing the implementation of replace_at to use rewriter buy us anything in terms of performance? Because otherwise we'll have a match statement regardless.

srk/src/arraylift.ml
nikhilpim: We still add to symbol_list whenever we pass over an integer existential or process an or (for the new branch variable). But since we have a mutable list in scope anyways, I will still factor out symbol_list.

srk/src/arraylift.ml
nikhilpim: That test case was unproblematic - after applying map_elim it was still UNSAT.

Because our algorithm automatically assumes a universal at any atom, we treat the above test case as exists n,m : int. exists a : array. forall i. n = m /\ a[n] != a[m].

The result of calling map elim looks like:

exists b, n, m, a_n, a_m. forall i. exists a_i . n = m /\ ((b=0 /\ a_n < a_m /\ FC) \/ (b = 1 /\ a_m < a_n /\ FC))

where:

FC: (m = i => a_m = a_i) /\ (n = i => a_n = a_i)

And this formula is UNSAT as expected.

srk/src/arraylift.ml
zkincaid: My understanding is that the purpose of `replace`/`replace_at` is to replace every array selection term according to `replace_access` (but the reason they're two separate functions is that `replace` operates on formulas and `replace_at` operates on terms).  If that's the case,
`fun expr -> rewrite srk ~up:(function Select(arr, index) -> replace_access arr index env qo | x -> x) expr`
does the same thing, and works for both formulas and terms.

srk/src/arraylift.ml
zkincaid: Ah, I see -- atomic formulas are treated as being universally quantified per line 140.  But then shouldn' this be treated the same as

exists n,m : int. exists a : array. (forall i. n = m) /\ (forall i. a[n] != a[m]).

(which should also give the incorrect result?). 

Is the existentially-quantified "b"  variable really introduced for this example?  It shouldn't be -- if the formula is unsat because it's asserting both b=0 and b=1, then it's masking the problem.

nikhilpim: Yes, it does get treated as (forall i. n = m) /\ (forall i. a[n] != a[m]), but this gets "bubbled" to forall i. (n = m /\ a[n] != a[m]) per one of the rewrite rules. It is true that this results in the functional consistency axioms being duplicated in each conjuct.

The FC axioms need to appear inside the forall, right? Because the universal "i" is needed within the definition of FC.

The existential b gets introduced because a[n] != a[m] gets rewritten to a[n] < a[m] \/ a[n] > a[m] to get rid of the negation.

srk/src/arraylift.ml
nikhilpim: That is true, but it is also responsible for replacing existential variables with the new skolem constants that were introduced in `go`. The logic for this is in the Var case of replace_at and the Exists case of replace. This is difficult to implement as a rewriter because we need additional context about the quantifier offset (`qo`) of variables we should replace versus leave unchanged.

nikhilpim:
@zkincaid Finished changes as discussed in meeting.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,192 @@
...
+    | `Not _ -> failwith "Not is not supported in Skolem fragment"
+  in
+  let forall, retf = go form Syntax.Env.empty in 
+  (!leading_existentials, (if forall then (Some forall_symbol) else None), retf)
+
+
+
+(* array_exponentiate lifts exponentiation operator e to work over formulas with arrays. *)
+let array_exponentiate (srk : 'a context) (e : 'a exp_op) : ('a TransitionFormula.t -> 'a Syntax.arith_term -> 'a Syntax.formula) = 
+  let e' (tf : 'a TransitionFormula.t) (k : 'a arith_term) : ('a formula) =
```
zkincaid: Remove the definition of e'?

srk/src/arraylift.ml
```diff
@@ -0,0 +1,192 @@
...
+(* array_exponentiate lifts exponentiation operator e to work over formulas with arrays. *)
+let array_exponentiate (srk : 'a context) (e : 'a exp_op) : ('a TransitionFormula.t -> 'a Syntax.arith_term -> 'a Syntax.formula) = 
+  let e' (tf : 'a TransitionFormula.t) (k : 'a arith_term) : ('a formula) = 
+    let f = TransitionFormula.formula tf in 
+    let array_symbols = List.filter (fun (s, s') -> typ_symbol srk s = `TyArr && typ_symbol srk s' = `TyArr) (TransitionFormula.symbols tf) in 
+    let zs = List.fold_left (fun acc (s, s') -> (s, mk_symbol srk `TyInt) :: (s', mk_symbol srk `TyInt) :: acc) [] array_symbols in 
+    let j = mk_symbol srk ~name:"j" `TyInt in 
+    let j' = mk_symbol srk ~name:"j'" `TyInt in 
+    let projected_formula = mk_and srk (f :: (mk_eq srk (mk_const srk j) (mk_const srk j')) :: (List.map (fun (a, z) -> mk_eq srk (mk_select srk (mk_const srk a) (mk_const srk j)) (mk_const srk z)) zs)) in 
+    let projected_formula = List.fold_left (fun acc (s, s') ->
```
zkincaid: TransitionFormula.symbols includes symbols corresponding to unprimed/primed "program variables", but not Skolem constants -- array-sorted Skolem constants also need to be removed.  Can do this `Syntax.symbols f` should give you what you need.

srk/src/arraylift.ml
```diff
@@ -0,0 +1,192 @@
...
+      let acc = if typ_symbol srk s = `TyArr then mk_exists_const srk s acc else acc in 
+      let acc = if typ_symbol srk s' = `TyArr then mk_exists_const srk s' acc else acc in 
+      acc
+      ) projected_formula (TransitionFormula.symbols tf) in 
+
+    let existentials, forall, skol = bubble_sym srk projected_formula in
+    let with_universal = match forall with
```
zkincaid: I'm not confident in the logic here.  A formula that starts without a forall might need one in the end (since we can introduce functional consistency axioms that talk about the universally quantified variable, which will be left as a skolem constant if we don't apply mk_forall_const).

zkincaid: With that in mind -- does bubble_sym need to compute that middle "forall" value?

nikhilpim: This is true - we don't need to do any computation for the forall value, it will always exist and it will always be the forall_symbol instantiated at the top of bubble_sym. I changed bubble_sym to just return this value, and removed forall from the internal go loop.

